Issue #269, opened 2015-01-02
Error in clojure-indent-line at start of file


j0ni:
When I hit return at the start of the first line in a clojure file, I see this:

```
clojure-indent-line: Args out of range: 0, 0
```

I'm using current master clojure-mode installed via melpa, and emacs 24.4.

PR #272, opened 2015-01-12
Fix broken `indent-for-tab-command' when point is 1

abo-abo:
* clojure-mode.el (clojure-in-docstring-p): Don't call
`(get-text-property 0)`, since it will throw.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## master (unreleased)
 
+### Bugs fixed
+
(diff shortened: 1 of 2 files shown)


bbatsov:
Looks like a fix for https://github.com/clojure-emacs/clojure-mode/issues/269. You might mention this in the commit message - "[Fix #269] ...". You might also mention this in the changelog.

abo-abo:
I made the updates that you asked. I'm not sure you noticed, since I used `push --force`.

bbatsov:
:+1: Thanks!

PR #281, opened 2015-03-20
Support prismatic schema s/def*.

tiensonqin:
We use schema.core/defn a lot, even more than clojure.core/defn.

### DIFF
--- a/clojure-mode.el
+++ b/clojure-mode.el
@@ -334,7 +334,7 @@ If JUSTIFY is non-nil, justify as well as fill the paragraph."
 (defun clojure-match-next-def ()
   "Scans the buffer backwards for the next \"top-level\" definition.
 Called by `imenu--generic-function'."
-  (when (re-search-backward "^(def\\sw*" nil t)
+  (when (re-search-backward "^([s/]*def\\sw*" nil t)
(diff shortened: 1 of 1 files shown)


expez:
I've copied `clojure-match-next-def` into my own `.emacs` for this reason too.  Next time we should probably pull this regexp into a var, but I think this is OK as 99.9% of all schema users alias it as `s`.

bbatsov:
I don't like this solution. We should look for any namespace prefixed `def`-something, instead of hardcoding `s/`. Same thing we do for font-locking.

tiensonqin:
I know it's not beautiful, this is more of suggestion. Maybe we could make
the regex to a var that user can reset.
On Mar 20, 2015 8:06 PM, "Bozhidar Batsov" <notifications@github.com> wrote:
>
> I don't like this solution. We should look for any namespace prefixed
def-something, instead of hardcoding s/. Same thing we do for font-locking.

bbatsov:
> def-something, instead of hardcoding s/. Same thing we do for font-locking.

That'd be an overkill IMO. We just need a regexp that's generic - e.g. `any-ns/defanything`.

Issue #271, opened 2015-01-11
Create indentation tests


bbatsov:
Such tests would be really useful and will make the code less vulnerable to regressions. 

We can borrow ideas from [elixir-mode](https://github.com/elixir-lang/emacs-elixir/blob/master/test/elixir-mode-indentation-test.el) and [swift-mode](https://github.com/chrisbarrett/swift-mode/blob/master/test/indentation-tests.el).

andrewhr:
@bbatsov looks interesting. One idea though: since indentation is _idempotent_, the test macro could receive only the expected version. That way the tests could read a bit more like documentation. Something like:

```elisp
(def-indentation-test vertically-aligns-function-args-spanning-multiple-lines 
  "
(filter even?
        (range 1 10))
")
```

Yep, I stealed the example from your style guide :wink:

the-kenny:
You could even read the forms from an external file - no need to embed code in strings which always seems messy.

PR #283, opened 2015-04-01
Support prefixed syms in clojure-indent-function

Chouser:
When a Clojure lib uses the same var name as clojure.core, it may have prefer different indent rules. This change allows specific prefixed symbols to override the indent rule of the unprefixed symbol.

For example, you can have a rule like (do 0) as well as (my-ns/do 1) and the latter will be preferred for actual uses of my-ns/do.

### DIFF
--- a/clojure-mode.el
+++ b/clojure-mode.el
@@ -685,7 +685,8 @@ This function also returns nil meaning don't specify the indentation."
              (function-tail (first
                              (last
                               (split-string (substring-no-properties function) "/")))))
-        (setq method (get (intern-soft function-tail) 'clojure-indent-function))
+        (setq method (or (get (intern-soft function) 'clojure-indent-function)                                                                                                                                      
+                         (get (intern-soft function-tail) 'clojure-indent-function)))                                                                                                                               
         (cond ((member (char-after open-paren) '(?\[ ?\{))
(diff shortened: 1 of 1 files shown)


bbatsov:
Looks good. Can you also update the changelog and add a couple of basic tests for this? (see https://github.com/clojure-emacs/clojure-mode/issues/271 for details)

Chouser:
Sure.

It looks like there's no code in the repo for indentation tests yet? Were you hoping I'd write that too? My elisp-foo isn't very strong, but if that was your intention I could take a swing at it.

bbatsov:
> It looks like there's no code in the repo for indentation tests yet? Were you hoping I'd write that too? My elisp-foo isn't very strong, but if that was your intention I could take a swing at it.

Just copy something from the modes I mentioned. The "framework" code for the indentation tests should be pretty much the same for any mode.

bbatsov:
@Chouser I added some basic tests to get you started, so I guess we can wrap finish this PR.

expez:
@Chouser ping

bbatsov:
I'll merge this and add the tests myself.

Chouser:
@bbatsov Sorry I dropped the ball on this; thanks for picking it up.

PR #291, opened 2015-04-29
Fix indentation of specify and specify!

weavejester:
When using `specify` or `specify!`, clojure-mode currently fails with:

    clojure-indent-function: Wrong type argument: number-or-marker-p, (2)

This seems to be because those functions have a clojure-backtracking-indent without an associated clojure-indent defined, as is the case with `extend-type` and `extend-protocol`, which have the same indentation scheme.

This PR fixes the issue.


bbatsov:
:+1:

PR #299, opened 2015-06-11
[Fix #295] clojure-in-docstring-p not using point

### LINKED ISSUE #295: clojure-in-docstring-p
There is an off-by-one error in clojure-in-docstring-p.

It happens when the first line is zero width!

Debugger entered--Lisp error: (args-out-of-range 0 0)
  get-text-property(0 face)
  clojure-in-docstring-p()
  (if (clojure-in-docstring-p) (save-excursion (beginning-of-line) (if (looking-at "^\\s-*") (progn (replace-match (clojure-docstring-fill-prefix))))) (lisp-indent-line))
  clojure-indent-line()
  indent-according-to-mode()
  electric-indent-post-self-insert-function()
  self-insert-command(1)
  newline(nil 1)
  call-interactively(newline nil nil)
  command-execute(newline)

expez:
The problem in #295 is caused by us checking the text property of the
character penultimate character in the current line.  When point is at
the beginning of the buffer we're asking for the property at position 0,
which doesn't exist.

The solution is to not worry about what the text property is for the
penultimate character of the current line and instead of check what it
is at the position of point.  This has the benefit of aligning the
function result, name and docstring :)

### DIFF
--- a/clojure-mode.el
+++ b/clojure-mode.el
@@ -275,9 +275,7 @@ ENDP and DELIMITER."
 
 (defsubst clojure-in-docstring-p ()
   "Check whether point is in a docstring."
-  (unless (bobp)
-    (eq (get-text-property (1- (point-at-eol)) 'face)
-        'font-lock-doc-face)))
+  (eq (get-text-property (point) 'face) 'font-lock-doc-face))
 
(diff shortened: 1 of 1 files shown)


bbatsov:
Sounds reasonable.

PR #306, opened 2015-07-16
Fixed font-locking for def* with support of special chars

BrunoBonacci:
changed the regex to match "def" + anything until the first space " ",
this support font locking for things like `defabc*`, `defyyy!`, etc

### DIFF
--- a/clojure-mode.el
+++ b/clojure-mode.el
@@ -380,7 +380,7 @@ Called by `imenu--generic-function'."
        (2 font-lock-type-face nil t))
       ;; Function definition (anything that starts with def and is not
       ;; listed above)
-      (,(concat "(\\(?:[a-z\.-]+/\\)?\\(def[^ ]*\\)"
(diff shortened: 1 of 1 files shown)


bbatsov:
That's a reasonable change. Before we can get this merged a few things need to be done:

* add unit tests
* add a changelog entry
* squash all commits into one
* use a commit message that matches our contribution guidelines

bbatsov:
@BrunoBonacci ping :-)

BrunoBonacci:
Hi,
I'm on holidays until mid of next week. I will take a look to it when I'm back. 
A disclaimer though: I'm a Clojure developer but I don't have any elisp experience so I might need some pointers. Anyway I'll give it a try. 

Bruno

bbatsov:
For the unit tests - we already have a bunch of font-locking tests, so just take a look at them and a few more covering your change.

Issue #311, opened 2015-07-29
Automatically created test namespace is wrong when nested source dir


yatesco:
Assume I have `src/cljc/a/b/c.cljc`. If I navigate in dired to test/cljc/a/b and then C-x C-f new_test.cljc then a new buffer is helpfully created `(ns cljc.a.b.new_test)` (filed in #310). However, it also requires:

```Clojure
(:require [cljc.a.b.new :refer :all]
            [clojure.test :refer :all])
```

(In terms of style I would suggest it is actually `[a.b.new :as sut]` for consistency, but that is a personal style issue).

Namespaces are setup correctly so `src/cljc` is a source path and `test/cljc` is a test path according to project.clj.

Libraries are Clojure 1.7.0 and snapshot cider.

bbatsov:
Hmm, this definitely not in `clojure-mode`. @expez I'm assuming this is coming from `clj-refactor`, right?

P.S. I also dislike the usage of `:refer :all`.

expez:
The broken function is `clojure-expected-ns` which is the cause of this: https://github.com/clojure-emacs/clojure-mode/issues/310#issuecomment-125926167

`clj-refactor` uses that same function to create the test namespace which prompted this issue to be created

expez:
This can be closed too, was hoping github would understand multiple issues closed by same commit

yatesco:
Thanks for the quick response guys.

PR #312, opened 2015-07-29
[Fix #310,#311] clojure-expected-ns with src/cljc

### LINKED ISSUE #310: Wrong namespace when source path is src/cljc
Assume I have `src/cljc/a/b/c.cljc`. If I navigate in dired to `src/cljc/a/b` and then `C-x C-f new.cljc` then a new buffer is created `(ns cljc.a.b.new)`.

Namespaces are setup correctly so `src/cljc` is a source path according to project.clj.

Libraries are Clojure 1.7.0 and snapshot cider.

expez:
When the source path is src/{clj,cljc,cljs} instead of just src/
clojure-expected ns would create namespaces like clj.my-project.my-ns
whereas what's wanted is my-project.my-ns.

Reading boot.clj or project.clj to find out the user's src dirs is out
of scope for clojure-mode, so we use the simply heuristic that no
namespace should start with clj, cljc or cljs because these are the
idiomatic source directories in multi-source projects.

When improving clojure-expected-ns I extracted out two utilities,
clojure-project-dir and clojure-project-relative-path.  These
utilities already exist in clj-refactor so I opted to make them public
rather than private, as they are generally useful.

### DIFF
--- a/clojure-mode.el
+++ b/clojure-mode.el
@@ -979,18 +979,31 @@ nil."
 
 
 
-(defun clojure-expected-ns ()
-  "Return the namespace name that the file should have."
-  (let* ((project-dir (file-truename
-                       (or (locate-dominating-file default-directory
-                                                   "project.clj")
-                           (locate-dominating-file default-directory
-                                                   "build.boot"))))
-         (relative (substring (file-truename (buffer-file-name))
-                              (length project-dir)
-                              (- (length (file-name-extension (buffer-file-name) t))))))
-    (replace-regexp-in-string
-     "_" "-" (mapconcat 'identity (cdr (split-string relative "/")) "."))))
+(defun clojure-project-dir ()
+  "Return the absolute path to the project's root directory."
+  (file-truename
+   (or (locate-dominating-file default-directory
+                               "project.clj")
+       (locate-dominating-file default-directory
+                               "build.boot"))))
+
+(defun clojure-project-relative-path (path)
+  "Denormalize PATH by making it relative to the project root."
+  (file-relative-name path (clojure-project-dir)))
+
+(defun clojure-expected-ns (&optional path)
+  "Return the namespace matching PATH.
+
+PATH is expected to be an absolute file path.
+
+If PATH is nil, use the path to the file backing the current buffer."
+  (let* ((relative (clojure-project-relative-path
+                    (or path (file-truename (buffer-file-name)))))
+         (sans-file-type (substring relative 0 (- (length (file-name-extension path t)))))
+         (sans-file-sep (mapconcat 'identity (cdr (split-string sans-file-type "/")) "."))
+         (sans-underscores (replace-regexp-in-string "_" "-" sans-file-sep)))
+    ;; Drop prefix from ns for projects with structure src/{clj,cljs,cljc}
+    (replace-regexp-in-string "\\`clj[scx]?\\." "" sans-underscores)))
 
 (defun clojure-insert-ns-form-at-point ()
   "Insert a namespace form at point."
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 ### Bugs fixed
 
+* [#310](https://github.com/clojure-emacs/clojure-mode/issues/310) and [#311](https://github.com/clojure-emacs/clojure-mode/issues/311) Fix `clojure-expected-ns` in multi-source projects.
 * [#307](https://github.com/clojure-emacs/clojure-mode/issues/307): Don't highlight `handle` and `handler-case` as keywords.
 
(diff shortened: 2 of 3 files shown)


clojure-mode.el
```diff
@@ -979,18 +979,31 @@ nil."
...
-                       (or (locate-dominating-file default-directory
-                                                   "project.clj")
-                           (locate-dominating-file default-directory
-                                                   "build.boot"))))
-         (relative (substring (file-truename (buffer-file-name))
-                              (length project-dir)
-                              (- (length (file-name-extension (buffer-file-name) t))))))
-    (replace-regexp-in-string
-     "_" "-" (mapconcat 'identity (cdr (split-string relative "/")) "."))))
+(defun clojure-project-dir ()
```
bbatsov: Turns out this implementation is problematic. See https://github.com/clojure-emacs/cider/pull/1215

clojure-mode.el
```diff
@@ -979,18 +979,31 @@ nil."
...
+  "Return the absolute path to the project's root directory."
+  (file-truename
+   (or (locate-dominating-file default-directory
+                               "project.clj")
+       (locate-dominating-file default-directory
+                               "build.boot"))))
+
+(defun clojure-project-relative-path (path)
+  "Denormalize PATH by making it relative to the project root."
+  (substring path (length (clojure-project-dir))))
```
bbatsov: See http://www.gnu.org/software/emacs/manual/html_node/elisp/Relative-File-Names.html

bbatsov: Generally using substring for file path manipulation is a bad idea.

clojure-mode.el
expez: This fix solves the problematic namespaces for 99.9% of affected projects. Let's solve one problem at a time?

clojure-mode.el
expez: Yeah, this makes sense.  In my defense I didn't write that code, I just moved it out into a utility function :p 

Thanks for the link to the emacs manual, I had no idea that function existed.

clojure-mode.el
bbatsov: I mentioned this mostly because if we have the same function here, there'll be little point in fixing the problem in cider. Anyways, that's not terribly important right now.

clojure-mode.el
```diff
@@ -979,18 +979,31 @@ nil."
...
+PATH is expected to be an absolute file path.
+
+If PATH is nil, use the path to the file backing the current buffer."
+  (let* ((relative (clojure-project-relative-path
+                    (or path (file-truename (buffer-file-name)))))
+         (sans-file-type (substring relative 0 (- (length (file-name-extension path t)))))
+         (sans-file-sep (mapconcat 'identity (cdr (split-string sans-file-type "/")) "."))
+         (sans-underscores (replace-regexp-in-string "_" "-" sans-file-sep)))
+    ;; Drop prefix from ns for projects with structure src/{clj,cljs,cljc}
+    (replace-regexp-in-string "\\`\\(clj.\\|cljs.\\|cljc.\\)" "" sans-underscores)))
```
bbatsov: clj[sc]?\\.

Not sure if something isn't needed for cljx as well.

expez: Updated with `cljx` support too

clojure-mode.el
```diff
@@ -979,18 +979,31 @@ nil."
...
+PATH is expected to be an absolute file path.
+
+If PATH is nil, use the path to the file backing the current buffer."
+  (let* ((relative (clojure-project-relative-path
+                    (or path (file-truename (buffer-file-name)))))
+         (sans-file-type (substring relative 0 (- (length (file-name-extension path t)))))
+         (sans-file-sep (mapconcat 'identity (cdr (split-string sans-file-type "/")) "."))
+         (sans-underscores (replace-regexp-in-string "_" "-" sans-file-sep)))
+    ;; Drop prefix from ns for projects with structure src/{clj,cljs,cljc}
+    (replace-regexp-in-string "\\`clj[scx]?." "" sans-underscores)))
```
bbatsov: Is `.` here for any character or this an oversight?

bbatsov:
Mention those 2 bugfixes in the changelog as well.

clojure-mode.el
expez: The bad ns look like `clj.foo.bar.baz` if we don't remove the `.` after `clj` then that will be the first character in the ns.

bbatsov: That's why I asked - `.` will match **any** character. You need to escape it.

expez: done

expez:
> Mention those 2 bugfixes in the changelog as well.

done

bbatsov:
Everything looks solid now. Thanks!

PR #313, opened 2015-07-30
Fix for error in `clojure-expected-ns`.

vmfhrmfoaj:
When call `clojure-expected-ns` function without arguments,
error occurred.

related commit: 4be6843

### DIFF
--- a/clojure-mode.el
+++ b/clojure-mode.el
@@ -997,8 +997,8 @@ nil."
 PATH is expected to be an absolute file path.
 
 If PATH is nil, use the path to the file backing the current buffer."
-  (let* ((relative (clojure-project-relative-path
-                    (or path (file-truename (buffer-file-name)))))
+  (let* ((path (or path (file-truename (buffer-file-name))))
+         (relative (clojure-project-relative-path path))
          (sans-file-type (substring relative 0 (- (length (file-name-extension path t)))))
          (sans-file-sep (mapconcat 'identity (cdr (split-string sans-file-type "/")) "."))
          (sans-underscores (replace-regexp-in-string "_" "-" sans-file-sep)))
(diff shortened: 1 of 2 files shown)


clojure-mode.el
```diff
@@ -997,8 +997,8 @@ nil."
 PATH is expected to be an absolute file path.
 
 If PATH is nil, use the path to the file backing the current buffer."
-  (let* ((relative (clojure-project-relative-path
-                    (or path (file-truename (buffer-file-name)))))
+  (let* ((path (or path (file-truename (buffer-file-name))))
+         (relative (clojure-project-relative-path path))
          (sans-file-type (substring relative 0 (- (length (file-name-extension path t)))))
```
expez: If you can't spot the problem, @bbatsov, `path` will be `nil` on this line when the function is called without any arguments.

expez:
Thanks for fixing this.

Testing this function was surprisingly hard due to the myriad calls against some current state in emacs.  Ironically this bug was introduced when I refactored the function to make it easier to test :/

test/clojure-mode-util-test.el
```diff
@@ -42,7 +42,16 @@
     (cl-letf (((symbol-function 'clojure-project-relative-path)
                (lambda (&optional current-buffer-file-name)
                  project-relative-clj-file-path)))
-      (should (string= (clojure-expected-ns clj-file-path) clj-file-ns)))))
+      (should (string= (clojure-expected-ns clj-file-path) clj-file-ns))))
+
+  (ert-deftest expected-ns-2 ()
```
bbatsov: You can use a descriptive name for the test. :-)

bbatsov:
Drop the `.` from the commit message's title.

vmfhrmfoaj:
I was changed a commit message and test name.

bbatsov:
:+1: Thanks!

PR #316, opened 2015-08-03
Fixed font-locking for def* with support of special chars

BrunoBonacci:
changed the regex to match "def" + anything until the first space " " (or blank),
this support font locking for things like `defabc*`, `defyyy!`, etc.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
 * [#302](https://github.com/clojure-emacs/clojure-mode/pull/302): Add new sexp navigation commands. `clojure-forward-logical-sexp` and `clojure-backward-logical-sexp` consider `^hints` and `#reader.macros` to be part of the sexp that follows them.
 * [#303](https://github.com/clojure-emacs/clojure-mode/issues/303): Handle `boot` projects in `clojure-expected-ns`.
 
+### Bugs fixed
+
+* Fix font-locking for def with special chars such as: `defn*`, `defspecial!`.
+
 ## 4.1.0 (20/06/2015)
 
 ### Changes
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@
 * [#302](https://github.com/clojure-emacs/clojure-mode/pull/302): Add new sexp navigation commands. `clojure-forward-logical-sexp` and `clojure-backward-logical-sexp` consider `^hints` and `#reader.macros` to be part of the sexp that follows them.
 * [#303](https://github.com/clojure-emacs/clojure-mode/issues/303): Handle `boot` projects in `clojure-expected-ns`.
 
+### Bugs fixed
+
+* Fix font-locking for def with special chars such: `defn*`, `defspecial!`
```
bbatsov: such as

This sentence should end with with a `.`.

bbatsov:
The commit message should say "Handle properly def* symbols, containing special chars" or something like it. We don't use past tense for commit message titles.

clojure-mode.el
```diff
@@ -380,7 +380,7 @@ Called by `imenu--generic-function'."
        (2 font-lock-type-face nil t))
       ;; Function definition (anything that starts with def and is not
       ;; listed above)
-      (,(concat "(\\(?:[a-z\.-]+/\\)?\\(def\[a-z\-\]*-?\\)"
+      (,(concat "(\\(?:[a-z\.-]+/\\)?\\(def[^ \r\n\t]*\\)"
```
bbatsov: Isn't this the same as using `[:space:]`?

BrunoBonacci: Not really... I'm using `^` for exclusion.
So I'm saying take anything that starts with a `def` and it is followed by anything which is **NOT** a blank 0 or more times. So to be fair should be something like `[^[:space:]]`.
Since different regex engines use different character classes (`[:space:]`, `[:blank:]`, `[:s:]`, `\s`) I wasn't sure about the correct one for elisp and I took reference from line 387 and 389 (just below)

bbatsov:
:+1: Thanks!

PR #318, opened 2015-09-07
Font-lock `code` marks, just like `elisp-quotes'

Malabarba:


bbatsov:
I've been thinking of adding this, but I'm not sure this simple solution is robust enough. Shouldn't we actually do some additional processing of docstrings? I believe this is how it's implemented in `lisp-mode` and friends.

Malabarba:
I dunno, but the version here works fine in docstrings and comments. I'll try to see how it's done in elisp.

bbatsov:
Guess I had too much faith in `lisp-mode`. I checked their solution and it's actually not considering the context at all - meaning `fdfdf' is font-locked the same regardless of where it appears (even at the top-level). If I knew this I would have added such font-locking eons ago.  :-)

Malabarba:
:smile:

PR #320, opened 2015-09-10
Define a variable for locally customizing indentation

Malabarba:


Malabarba:
The first commit merged the two symbol properties into one.
I did it because it will make things simpler on the CIDER side, and I think I did it in a backwards compatible way. Let me know if you disagree.

Malabarba:
Ok, I see now that merging the two like this is not backwards compatible.
Could you explain to me how the backtracking spec works? I'll need to understand it if I'm to merge the two properly.

bbatsov:
> Could you explain to me how the backtracking spec works? I'll need to understand it if I'm to merge the two properly.

Wish I could. :-) This is where we see the failure of lack of documentation and tests.

bbatsov:
Btw, I wrote a bit more about this on gitter. This whole functionality has a funny story - it was introduced in the very first version of `clojure-mode` in 2008 and has been changed little since. The only big change is that at some point someone asked for it to be enabled by default. The only documentation is the code (which at some point I was superficially familiar with) - no docs, no tests, no nothing. I remember I figured out the workings at some point mostly by looking at the default config (alas this knowledge has faded by now and I don't have time to rehash it). Frankly, I'd rather dispose of it and have just one indentation function which understands more complex indentation specifications by default. 

You have carte blanche to deal with this as you see fit, as long we simplify the inner workings in the process.

Malabarba:
Ok, I may have abused your carte blanche here.
Rewrote the backtracking code from scratch, and changed quite a bit of the regular code as well.
It also relies a bit less on lisp-mode, which fixed another couple of bugs.
The new form of indent specs is described in the commit message. I'll document it, of course, but I want to hear the opinions of other people first, including the clojure-list.

bbatsov:
> The new form of indent specs is described in the commit message. I'll document it, of course, but I want to hear the opinions of other people first, including the clojure-list.

Haven't went through the changes in great detail, but they look good to me. You'll definitely won't get much feedback about the quality of new code on the Clojure mailing list, but might receive some comments about the structure of the indentation spec at least.

Malabarba:
>but might receive some comments about the structure of the indentation spec at least

Oh yes, that's what I meant.

Malabarba:
Fixed the docstring and added some tests for `normal-indent` commit.

The other commit didn't actually change any functionality. So the existing tests should be a great way to verify it's validity. :-)

bbatsov:
> The other commit didn't actually change any functionality. So the existing tests should be a great way to verify it's validity. :-)

There were zero tests for the backtracking indent... :-) Adding tests in an unavoidable task at this point.

Malabarba:
Ok. I've added a couple of tests and I think I'm ready to merge this.

bbatsov:
On a related note - the section in the README about indentation has to be updated/extended.

PR #321, opened 2015-09-15
[Fix clojure-emacs/cider#1323] Sexp navigation near end of buffer

Malabarba:


clojure-mode.el
```diff
@@ -1087,8 +1088,7 @@ Returns a list pair, e.g. (\"defn\" \"abc\") or (\"deftest\" \"some-test\")."
 (defun clojure--looking-at-logical-sexp ()
   "Return non-nil if sexp after point represents code.
 Sexps that don't represent code are ^metadata or #reader.macros."
-  (forward-sexp 1)
-  (forward-sexp -1)
+  (comment-forward (point-max))
```
expez: Should you be calling `comment-normalize-vars` first here?

Malabarba: Good catch. Fixed.

bbatsov:
I guess more changes would be needed.

Malabarba:
What do you mean?

expez:
@Malabarba the tests are failing

Malabarba:
Oh. :)

Malabarba:
Ah, it's some weird interaction with the current indentation engine. It's going to fix itself when we merge the other PR, so we can just sit on this one until then. :)

bbatsov:
OK

bbatsov:
Can you, please, rebase this?

Malabarba:
Done. Added some tests too.

PR #322, opened 2015-09-16
Add clojure-in-string-p and clojure-in-comment-p

expez:
I needed these when working on CIDER, but I think they belong here
alongside clojure-in-docstring-p.

### DIFF
--- a/clojure-mode.el
+++ b/clojure-mode.el
@@ -638,6 +638,14 @@ point) to check."
           (replace-match (clojure-docstring-fill-prefix))))
     (lisp-indent-line)))
 
+(defun clojure--symbol-get (function-name property)
+  "Return the symbol PROPERTY for the symbol named FUNCTION-NAME.
+FUNCTION-NAME is a string.  If it contains a `/', also try only the part after the `/'."
+  (or (get (intern-soft function-name) property)
+      (and (string-match "/\\([^/]+\\)\\'" function-name)
+           (get (intern-soft (match-string 1 function-name))
+                property))))
+
 (defun clojure-indent-function (indent-point state)
   "When indenting a line within a function call, indent properly.
 
@@ -670,7 +678,12 @@ This function also returns nil meaning don't specify the indentation."
           (if (not (> (save-excursion (forward-line 1) (point))
                       calculate-lisp-indent-last-sexp))
               (progn (goto-char calculate-lisp-indent-last-sexp)
-                     (beginning-of-line)
+                     (skip-chars-backward "[:blank:]")
+                     ;; We're done if we find the start of line,
+                     (while (and (not (looking-at-p "^"))
+                                 ;; or start of sexp.
+                                 (ignore-errors (forward-sexp -1) t))
+                       (skip-chars-backward "[:blank:]"))
                      (parse-partial-sexp (point)
                                          calculate-lisp-indent-last-sexp 0 t)))
           ;; Indent under the list or under the first sexp on the same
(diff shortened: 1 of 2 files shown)


expez:
This works as it should in the repl too because the repl sets `beginning-of-defun-function` to something sensible.

expez:
I don't know how to fix the `clojure-in-docstring-p` predicate easily because the `defn` form is incredibly complicated.

I propose we simply remove this predicate and replace it's use with `clojure-in-string-p`.  The only consequence of that is that you get the more clever filling when you hit `M-q` inside regular strings.

clojure-mode.el
```diff
@@ -56,6 +56,7 @@
 
 
 
+
```
bbatsov: Why the extra blank here?

clojure-mode.el
```diff
@@ -277,8 +278,26 @@ ENDP and DELIMITER."
   (add-hook 'paredit-mode-hook #'clojure-paredit-setup))
 
 (defsubst clojure-in-docstring-p ()
-  "Check whether point is in a docstring."
-  (eq (get-text-property (point) 'face) 'font-lock-doc-face))
+  "True if point is in a docstring."
```
bbatsov: Frankly, I prefer docstrings that start with some verb (e.g. `return`).

clojure-mode.el
```diff
@@ -277,8 +278,26 @@ ENDP and DELIMITER."
...
+  (eq (get-text-property (point) 'face) font-lock-doc-face))
+
+(defsubst clojure-in-string-p (&optional beg)
+  "True if point is in a string.
+
+BEG is how far back we should look to determine the context
+around point.  The default is the value returned by
+`beginning-of-defun'"
```
bbatsov: This is missing a `.`.

clojure-mode.el
```diff
@@ -277,8 +278,26 @@ ENDP and DELIMITER."
...
+  (let ((beg (or beg (save-excursion (beginning-of-defun) (point)))))
+    (nth 3 (parse-partial-sexp beg (point)))))
+
+(defsubst clojure-in-comment-p (&optional beg)
+  "True if point is in a comment.
+
+If BEG is how far back we should look to determine the context
+around point.  The default is the value returned by
+`beginning-of-defun'"
```
bbatsov: Same here.

clojure-mode.el
```diff
@@ -277,8 +278,26 @@ ENDP and DELIMITER."
...
+BEG is how far back we should look to determine the context
+around point.  The default is the value returned by
+`beginning-of-defun'"
+  (let ((beg (or beg (save-excursion (beginning-of-defun) (point)))))
+    (nth 3 (parse-partial-sexp beg (point)))))
+
+(defsubst clojure-in-comment-p (&optional beg)
+  "True if point is in a comment.
+
+If BEG is how far back we should look to determine the context
```
bbatsov: This sentence is not very clear IMO.

test/clojure-mode-util-test.el
```diff
@@ -53,6 +64,27 @@
                          (clojure-expected-ns))
                        clj-file-ns)))))
 
+(ert-deftest clojure-in-comment-p-test ()
+  :tags '(utils)
```
bbatsov: Probably there should be some negative test cases as well.

expez:
Somehow managed to close this PR by force-pushing to the wrong branch.  The ticker at the top claims I opened this from my master branch so that's where I pushed heh

I'll re-open this.

expez:
Unaddressed issue from previous PR: 

I don't know how to fix the clojure-in-docstring-p predicate easily because the defn form is incredibly complicated.

I propose we simply remove this predicate and replace it's use with clojure-in-string-p. The only consequence of that is that you get the more clever filling when you hit M-q inside regular strings.

PR #323, opened 2015-09-16
Add clojure-in-string-p and clojure-in-comment-p

expez:
I needed these when working on CIDER, but I think they belong here
alongside clojure-in-docstring-p.


expez:
Unaddressed issue from previous PR:

I don't know how to fix the clojure-in-docstring-p predicate easily because the defn form is incredibly complicated.

I propose we simply remove this predicate and replace it's use with clojure-in-string-p. The only consequence of that is that you get the more clever filling when you hit M-q inside regular strings.

expez:
These are now part of cider-util.el

PR #326, opened 2015-09-29
Font-lock namespaced keywords

Malabarba:


bbatsov:
We also need a test for this.

Malabarba:
Done. :+1:

test/clojure-mode-font-lock-test.el
```diff
@@ -101,7 +101,13 @@ POS."
 
 (ert-deftest clojure-mode-syntax-table/fontify-clojure-keyword ()
   :tags '(fontification syntax-table)
-  (should (eq (clojure-test-face-at 2 11 "{:something 20}") 'clojure-keyword-face)))
+  (should (equal (clojure-test-face-at 2 11 "{:something 20}") '(clojure-keyword-face))))
```
bbatsov: Why does this return a list now?

test/clojure-mode-font-lock-test.el
```diff
@@ -101,7 +101,13 @@ POS."
...
+
+(ert-deftest clojure-mode-syntax-table/fontify-namespaced-keyword ()
+  :tags '(fontification syntax-table)
+  (should (equal (clojure-test-face-at 2 2 "{:alias/some 20}")  '(clojure-keyword-face)))
```
bbatsov: This doesn't look right. Shouldn't `alias` be font-locked using the `font-lock-type-face`?

Malabarba: It should, and it is. :)
This test is testing the `:` character.

test/clojure-mode-font-lock-test.el
Malabarba: Because the keyword rule now uses the `append` option (see https://github.com/clojure-emacs/clojure-mode/pull/326/files#diff-50632fdcf67b6d3c8505a6aa071a2bd0R459). So font-lock-mode adds it as a list (even if it turns out to be the only face).

test/clojure-mode-font-lock-test.el
Malabarba: I've fine tuned this test now to cover the entire symbol.

ilshad: Guys, why? It's very inconvenient now and it's weird.

Namespace within keyword should always be highlighted with the same color as keyword are, as it always was here and in all other editors, including vim and intellij.

It's so inconvenient that I'm looking now for a hack, how can I revert this behavior back without changing clojure-mode version...

expez: > Guys, why?

For consistency with how other namespaced stuff is highlighted.

> It's very inconvenient now and it's weird.

This is just your opinion and it's very much influenced by having had namespaced keywords fontlocked in this manner for years.  If you'd never seen one before it would look perfectly reasonable.  In a few weeks time the new normal will be normal.

> It's so inconvenient that I'm looking now for a hack, how can I revert this behavior back without changing clojure-mode version...

You can copy in an old version of clojure-mode, which is to your liking, and either never update or cherry-pick changes you like.

bbatsov: I agree with @expez 100%. The change is for the better and is not going anywhere.

Malabarba: Same here. :+1:

PR #333, opened 2015-10-13
Improve clojure-find-def matching (support hyphens)

maio:
Custom def... macros might contain hyphens (e.g. defxxx-yyy).


bbatsov:
A matching font-lock test is a prerequisite for this PR to be accepted.

bbatsov:
My bad - I meant to say "does the font-locking code match this behaviour".

maio:
OK, thanks for the feedback. I'll check that.

bbatsov:
Great.

bbatsov:
I plan to release `clojure-mode` 5.0 pretty soon, so if you want the PR in it you should wrap it in a day or two.

maio:
Is this test sufficient?

maio:
I fixed the test.

PR #346, opened 2015-11-17
Form alignment

Malabarba:


Malabarba:
Closed as I'm not quite happy with the feature yet.

clojure-mode.el
```diff
@@ -629,6 +629,81 @@ point) to check."
 (put 'definline 'clojure-doc-string-elt 2)
 (put 'defprotocol 'clojure-doc-string-elt 2)
 
+;;; Vertical alignment
+(defcustom clojure-binding-forms-regexp (regexp-opt (mapcar #'symbol-name
```
bbatsov: I'd really wouldn't use a regexp for those. Why not just a list of symbols?

PR #347, opened 2015-11-17
Form alignment

Malabarba:
Much nicer now.
This hooks into indent-line, so that no extra commands are required. 

If the user wants, a command is provided as well. We could bind it to `C-c SPC` in analogy to `C-x SPC`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### New features
 
+* Vertically align sexps with `C-c SPC`. This can also be done automatically (as part of indentation) by turning on `clojure-align-forms-automatically`.
 * Indent and font-lock forms that start with `let-`, `while-` or `when-` like their counterparts.
 * Apply the `font-lock-comment-face` to code commented out with `#_`.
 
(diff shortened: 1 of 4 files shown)


bbatsov:
Btw, I'm guessing you can add this to `lisp-mode`/`elisp-mode` down the road. Plenty of people would like it.

bbatsov:
We should wrap this as well. I know that a bunch of people are quite excited about the feature.

Malabarba:
I know, it's on my list. But I'm still going through bug fixes for the cider release. :)

bbatsov:
> But I'm still going through bug fixes for the cider release. :)

Yeah, those are way more important in the short term for sure. :-)

bbatsov:
> We could bind it to C-c SPC in analogy to C-x SPC.

That sounds reasonable. I missed it when I was originally reading the description. Once this and the comment changes are finalized I'll probably release them as 5.1.

Malabarba:
PR updated.

Malabarba:
And (I should say) heavily changed. It now uses the `align-region` function, instead of aligning manually, though we still have to do quite a bit of work because our alignment is sexp-based, and not regexp based.

bbatsov:
:+1:

PR #348, opened 2015-11-18
let letxxx, whenxxx, whilexxx indentation like def.

gfZeng:


Malabarba:
The right way to do this would be to set the indent-spec for these macros to `:defn`. 

However, I think this something for the users to configure. I don't see any reason to make it default. It should only make a difference when you don't write the condition (or the binding vector) on the same line as the macro. 

Unless you've noticed some other quirk about it that shouldn't be happening. When current outcome are you unhappy about?

gfZeng:
I found many clojure lib has some useful `condition macro` and `binding macro`,
the indentation is not satisfyingly.

eg.
`let-ajax` is a binding macro come from [jayq](https://github.com/ibdknox/jayq)

I think clojure developer use `condition macro` and `binding macro` often.

Malabarba:
I see. So your suggestion is to generalise our handling of these macros. I thought you wanted to change how let and when-let were indented. 
That's surely something I can agree with, but this is not the best way to do it. I'll look into doing that next week.

gfZeng:
@Malabarba 
Yes, you suggestion is right. I will close this PR.
my many friends need this feature like me.
Expected your better way.
thank you!

PR #350, opened 2015-11-26
Make the ns regexp more permissive

expez:
Any valid symbol can be used to name an ns.  The current regexp would fail
to parse symbols ending in certain characters, e.g. `+`.

This closes clojure-emacs/cider#1433


clojure-mode.el
```diff
@@ -1009,7 +1009,7 @@ nil."
                     (one-or-more (any whitespace "\n")))
       ;; why is this here? oh (in-ns 'foo) or (ns+ :user)
       (zero-or-one (any ":'"))
-      (group (one-or-more (not (any "()\"" whitespace))) word-end)))
+      (group (one-or-more (not (any "()\"" whitespace))) symbol-end)))
```
bbatsov: Not sure how correct our symbol definition is. If memory serves - not very correct. This change definitely requires unit tests illustrating it actually solves something.

expez: Our symbol definition is inherited from elisp so it's at least better than using word characters.  I can't think of any notable differences in clojure's and elisps notion of what a symbol is.

bbatsov: The problem is that I tried using symbol instead of word semantics when I inherited the mode and everything broke, therefore my reservations... Using symbol semantics is the reasonable thing to do for pretty much everything, but I really things something was fucked up regarding them. I recall `lisp-mode`'s code was full of word references itself, which was odd. 

At any rate - a few unit tests won't hurt.

Malabarba: AFAICT, that regexp should be just `(1+ (any symbol word))`. But you're the voice of experience here @bbatsov.

bbatsov: That sounds reasonable. But on a general note @expez is right that there should be fewer word checks and symbols checks, as we're talking about symbols here.

Malabarba: Yes. I agree too.

bbatsov: Btw, most of this word stuff predates me becoming the maintainer of `clojure-mode`. I always wanted to redo it (and unify the regex syntax used in `clojure-mode`), but never had the time to do.

Malabarba: > AFAICT, that regexp should be just (1+ (any symbol word)). 

Correction, what I wrote is not valid rx. It would have to be `(1+ (or (syntax word) (syntax symbol)))`

expez:
* Updated the regexp to match @Malabarba suggestion.
* Moved the inline, commented out tests, into a test file
* Removed duplicate tests
* Expanded tests to cover the change made here
* Cleaned up a comment slightly

expez:
Actually, I've restored the regexp to my first suggestion, because

```el
(group (one-or-more (or (syntax symbol) (syntax word))))
```

fails on namespaces like this:

```
"(ns #^{:bar true} foo)"
```

Malabarba:
Ideally, each of those tests would also check that `(equal "foo" (match-string 4 s))`. 
But other than that it looks dandy. :+1:

bbatsov:
Apart from @Malabarba's suggestion - a changelog entry would be nice. No other remarks from me. 

P.S. We should turn more comments into specs. :-)

expez:
Aright let's go

bbatsov:
:+1:

PR #351, opened 2015-11-30
Fix font-locking: '#', '*', '_', '+' in namespace alias

Bost:
(M-x clojure-mode-display-version) gives me:
clojure-mode (version 5.0.1)


Bost:
Hmm I see my PR breaks ns-alias font locking somewhere else... :( Please let me fix that before applying this PR. Sorry

Bost:
The namespace alias syntax is a bit more complicated than I thought. I have to invest some more time in this problem.

PR #352, opened 2015-12-01
[Fix #343] Don't font-lock the `:` in keywords

### LINKED ISSUE #343: Change the face for : in keywords
We talked about this on gitter a while back, but the more I think of this the more I'm convinced we shouldn't put a face on this token. Much like the `/` in namespaced stuff.  The `:` isn't really part of the keyword, but a token for the reader.  

```clj
user=> (name :foo)
;;=> "foo"
```

If the return value here was `":foo"`, then that would mean that the `:` was really part of the keyword, but since it isn't, I think we should drop the keyword face here.

expez:
The `:` in keywords were font-locked with the keyword face because we
inherited the rules for font-locking keywords from Elisp.

In Elisp a keyword is a symbol which starts with the `:` character.
Furthermore, the predicates symbolp and keywordp return T for keywords
in Elisp.  This is different from how a keyword is defined, and how it
behaves, in Common-lisp and in Clojure.

In common-lisp the `:` is a hint to the reader that the symbol should be
interned in the keyword package.

Clojure doesn't have a keyword package, but when the reader consumes a
`:` it means that the following symbol is to be stored in the keyword
table, residing in `Keyword.java`.  Unlike in Elisp, the `:` is never
stored anywhere. It's just the way keywords print and syntactic sugar
for keyword creation.

Thus, unlike in elisp, the `:` is a token meant for the reader, and not
part of the keyword itself, and should not be font-locked with the
keyword face.

#### Before

kw-before

#### After

image

Even though it's barely noticeable, I think this has the benefit of looking a lot better, but that part is highly subjective and shouldn't be part of the decision to merge this or not.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### New features
 
+* Don't font-lock the `:` in keywords.
 * Indent and font-lock forms that start with `let-`, `while-` or `when-` like their counterparts.
 
 ### Bugs fixed
--- a/clojure-mode.el
+++ b/clojure-mode.el
@@ -463,7 +463,7 @@ Called by `imenu--generic-function'."
       ;; foo/ Foo/ @Foo/ /FooBar
       ("\\(?:\\<:?\\|\\.\\)@?\\([a-zA-Z][.a-zA-Z0-9$_-]*\\)\\(/\\)" (1 font-lock-type-face) (2 'default))
       ;; Constant values (keywords), including as metadata e.g. ^:static
-      ("\\<^?\\(:\\(\\sw\\|\\s_\\)+\\(\\>\\|\\_>\\)\\)" 1 'clojure-keyword-face append)
+      ("\\<^?::?\\(\\(\\sw\\|\\s_\\)+\\(\\>\\|\\_>\\)\\)" 1 'clojure-keyword-face append)
       ;; Java interop highlighting
       ;; CONST SOME_CONST (optionally prefixed by /)
       ("\\(?:\\<\\|/\\)\\([A-Z]+\\|\\([A-Z]+_[A-Z1-9_]+\\)\\)\\>" 1 font-lock-constant-face)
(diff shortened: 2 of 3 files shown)


Malabarba:
> shouldn't be part of the decision to merge this or not.

I think that “how nice” it looks should be a big part of decisions pertaining to font-lock. :-)

The primary purpose of syntax-coloring, after all, is to make code easier to read. Teaching internal details of the language is secondary to that.

If we are going to be strict about language internals, we should also apply the same logic to 
  * double quotes (the string delimiters are not part of the string), 
  * the backlash in character literals (the `\` in `\c` is just a hint to the reader), 
  * the `#` in reader tags (`#inst` corresponds to the `inst` entry of `default-data-readers`, and the `#` is not part of the name).

expez:
> The primary purpose of syntax-coloring, after all, is to make code easier to read. 

Sure, and I don't think this hurts readability, except perhaps in the degenerate one-char-keyword case.

I actually agree with the proposed changes (except for the double quotes), even if you don't actually think that's a good idea.  Not highlighting the double quotes would be weird simply because strings are ubiquitous and everyone else highlights the double quotes along with the string content.

bbatsov:
Let's leave this open for a while and see what others think about the proposed change. I'm leaning against it, as for most people this is just the keyword literal syntax, regardless of language internals.

Bost:
It should be visually clear that ```:``` ```"``` etc. are not a part of keywords, strings, etc. and the same time the code should be nicely font-locked. The point is how to achieve both (without much bicycle-shedding)?

sbelak:
If you look at the documentation for any clj project (even the core language docs), everyone writes keywords, as if : was a part of the keyword. For instance the doc for `require`:

> A flag is a keyword.
> Recognized flags: :reload, :reload-all, :verbose

I see syntax highlighting as helping communicate clearly (and at a glance), and by this point I think most people's mental model is that : is a part of the keyword (even if they rationally know that is not the case).

agoldhammer:
I agree with sbelak. It is a useful visual cue to highlight keywords, and it is natural to think of the colon as part of the keyword, regardless of language internals.

dgutov:
> double quotes (the string delimiters are not part of the string)

I think this is the best argument against the proposal's logic. Strings are indeed ubiquitous, and people are used to having them highlighted together with the delimiters. But so are e.g. escape sequences.

The "After" image does look nice, though.

bbatsov:
OK, seems there's no support for this change.

PR #354, opened 2015-12-07
[Fix #277] Apply font-lock-comment-face to (comment) and #_

### LINKED ISSUE #277: Apply font-lock-comment-face to (comment...) and #_... expressions
Maybe this feature was added by merging #83 but then it was removed?

Malabarba:


bbatsov:
As I noted in the original I have some doubts about applying this to `comment`, as the behaviour there's not exactly the same as with `#_` (e.g. `comment` returns a result `nil`). 

@expez What do you think?

clojure-mode.el
```diff
@@ -327,7 +327,44 @@ If JUSTIFY is non-nil, justify as well as fill the paragraph."
...
+           (start (match-beginning 1))
+           (state (syntax-ppss start)))
+      ;; inside string or comment?
+      (if (or (nth 3 state)
+              (nth 4 state))
+          (clojure--search-comment-macro-internal limit)
+        (goto-char start)
+        (clojure-forward-logical-sexp 1)
+        ;; Data for (match-end 1).
+        (setf (elt md 3) (point))
```
bbatsov: I've almost never seen `setf` in Emacs Lisp code (although it was quite popular in CL).

clojure-mode.el
```diff
@@ -327,7 +327,44 @@ If JUSTIFY is non-nil, justify as well as fill the paragraph."
         (do-auto-fill)))))
 
 
+;;; #_ comments font-locking
```
bbatsov: This comment disagrees a bit with actual code (as it affects `comment` as well).

bbatsov:
Two more remarks - this needs a changelog entry and maybe there should be some config option about the `comment` macro. Anyways, let's hear what @expez and possibly others would have to say about the change.

expez:
I don't think we should font-lock `comment` with comment face.  The only time I see the `comment` macro used is to provide example usage, or to put some code in the buffer which can easily be evaluated (e.g. with <kbd> C-c C-e</kbd>) while developing. In those two scenarios font-locking is quite beneficial.  

I don't have time right now, but I think it would be a good idea to run a code search on github to check out how people are using `comment`.

clojure-mode.el
Malabarba: I almost always use it to destructively change a sequence or struct. :-)
It's considerably more convenient than something like `(setcar (cdr (cdr (cdr md))) start)`.

Malabarba:
I'll remove the comment macro support and just add an example on the docstring about how to support it.

clojure-mode.el
bbatsov: Fair enough. :-) I just said it's pretty uncommon.

bbatsov:
Sounds good to me.

bbatsov:
:+1:

PR #373, opened 2016-03-26
Treat , as a non-logical sexpr character

### LINKED ISSUE #367: Vertical alignment doesn't treat , correctly
```clojure
(def ^:private start-element
  {:meta "^", :meta* "#^", :vector "[",       :map "{"
   :list "(", :eval "#=",  :uneval "#_",      :fn "#("
   :set "#{", :deref "@",  :reader-macro "#", :unquote "~"
   :var "#'", :quote "'",  :syntax-quote "`", :unquote-splicing "~@"})
```

indents with align-vertically to 

```clojure
(def ^:private start-element
  {:meta "^"  , :meta* "#^" , :vector        "["  ,       :map        "{"
   :list "("  , :eval  "#=" ,  :uneval       "#_" ,      :fn          "#("
   :set  "#{" , :deref "@"  ,  :reader-macro "#"  , :unquote          "~"
   :var  "#'" , :quote "'"  ,  :syntax-quote "`"  , :unquote-splicing "~@"})
```

since `,` is lexical whitespace, the `,`s should remain attached to the preceding forms.

arrdem:
This causes clojure-forward-logical-sexp to skip over , when for
instance computing vertical alignments.

Fixes #367

### DIFF
--- a/clojure-mode.el
+++ b/clojure-mode.el
@@ -1451,7 +1451,7 @@ Returns a list pair, e.g. (\"defn\" \"abc\") or (\"deftest\" \"some-test\")."
 Sexps that don't represent code are ^metadata or #reader.macros."
   (comment-normalize-vars)
   (comment-forward (point-max))
-  (looking-at-p "\\^\\|#[?[:alpha:]]"))
+  (looking-at-p "\\^\\|#[?[:alpha:]]\\|,"))
 
 (defun clojure-forward-logical-sexp (&optional n)
   "Move forward N logical sexps.
--- a/clojure-mode.el
+++ b/clojure-mode.el
@@ -1465,6 +1465,7 @@ This will skip over sexps that don't represent objects, so that ^hints and
       (while (> n 0)
         (while (clojure--looking-at-non-logical-sexp)
           (forward-sexp 1))
+        (skip-chars-forward ",")
         ;; The actual sexp
         (forward-sexp 1)
         (setq n (1- n))))))
--- a/test/clojure-mode-indentation-test.el
+++ b/test/clojure-mode-indentation-test.el
@@ -540,6 +540,12 @@ x
  :b {:a  :a
      :aa :a}}")
 
+(def-full-align-test trailing-commas
+  "{:a {:a  :a,
+     :aa :a},
+ :b {:a  :a,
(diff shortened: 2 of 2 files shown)


Malabarba:
Well, I guess that was easier than expected. :smile: 
Could also add a test for this in the alignment tests? It's very easy to write. Just copy one of the other tests in there, and replace the code string with an example that failed before but works now.

Malabarba:
And thanks a ton btw!

arrdem:
Yeah I'm honestly kinda surprised that this is all it took even in some fairly torturous case...

2016-03-25-204426_251x86_scrot

arrdem:
I don't think this is right however 
2016-03-25-204523_311x87_scrot

arrdem:
This is also pretty tortured....

2016-03-25-204641_213x114_scrot

arrdem:
Also isn't right... not sure what's going on here.

2016-03-25-205058_615x105_scrot

Malabarba:
Hmm. Now that I think about it... The current code probably assumes that non-logical sexps come *before* the logical sexp, which is not true for commas. 
I don't think we need to change that, though (commas aren't really sexps anyway). It should be enough to just `(skip-chars-forward ",")` after moving over a *logical* sexp. This won't cover some of the more tortuous cases you tried above, but it should cover the case of a single comma after a sexp, which is really the only one I'm concerned with.

arrdem:
Okay. These two together seem to do the job better.
2016-03-26-232339_241x65_scrot

Malabarba:
Merged! I also added a bit of code so that `clojure-align` will automatically cleanup excessive commas.

arrdem:
Did you test this at all for yourself? I'm concerned that it has fairly significant limitations from mucking this afternoon :disappointed:

Malabarba:
I did, but I also changed the code a bit, which might have overcome the limitations. Have you tried the version on master?

arrdem:
No I've been in hospital all day. I'll check it out tomorrow.

Malabarba:
:+1: Let me know.

arrdem:
LGTM. Thanks Artur!

2016-03-28-102731_583x97_scrot

PR #376, opened 2016-04-09
Remove clojure-defun-indents custom variable

harto:
This variable appears to be outdated; indentation options are now specified using `put-clojure-indent`, `define-clojure-indent` and `clojure-backtracking-indent`.

I don't know if removing this breaks BC, if so, I'm happy to simply document it as deprecated in the source.


harto:
Oh, I think I might've simply misunderstood how this variable is used.
I think at one point I was setting it via `.dir-locals.el`, but perhaps it only takes effect when setting it via `customize-variable`.

harto:
Sorry for the noise :grimacing:

PR #379, opened 2016-05-02
Add threading macros related refactorings

benedekfazekas:
- [x] The commits are consistent with our [contribution guidelines](../CONTRIBUTING.md)

Code is ported from clj-refactor.el. Originally was mainly the work of
Magnar Sveen (@magnars) and Alex Baranosky (@AlexBaranosky).

The code here does not use paredit and have minor adjustments
but should have the same feature set as the original, see
the related tests ported. Also clojure-emacs/clj-refactor.el#259
is fixed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,8 @@ specific `clojure-mode` release.**
     - [Indentation of function forms](#indentation-of-function-forms)
     - [Indentation of macro forms](#indentation-of-macro-forms)
   - [Vertical alignment](#vertical-alignment)
+- [Refactoring support](#refactoring-support)
+  - [Threading macros](#threading-macros-related-features)
 - [Related packages](#related-packages)
 - [REPL Interaction](#repl-interaction)
   - [Basic REPL](#basic-repl)
@@ -219,6 +221,32 @@ Leads to the following:
 This can also be done automatically (as part of indentation) by
 turning on `clojure-align-forms-automatically`. This way it will
 happen whenever you select some code and hit `TAB`.
+
+## Refactoring support
+
+The available refactorings were originally created and maintained by the clj-refactor.el team. The ones implemented in Elisp only are gradually migrated to Clojure mode.
+
+### Threading macros related features
+
+* Thread an other expression.
+
+Thread another form into the surrounding thread. Both `->>` and `->` variants are supported. See demonstration on the [clj-refactor.el Wiki](https://github.com/clojure-emacs/clj-refactor.el/wiki/cljr-thread).
+
+* Unwind a threaded expression.
+
+Supports both `->>` and `->`. See demonstration on the [clj-refactor.el Wiki](https://github.com/clojure-emacs/clj-refactor.el/wiki/cljr-unwind-thread).
+
+* Wrap in thread first (`->`) and fully thread.
+
+Introduce the thread first macro and rewrite the entire form. With a prefix argument do not thread the last form. See demonstration on the [clj-refactor.el Wiki](https://github.com/clojure-emacs/clj-refactor.el/wiki/cljr-thread-first-all).
+
+* Wrap in thread last (`->>`) and fully thread.
+
+Introduce the thread last macro and rewrite the entire form. With a prefix argument do not thread the last form. See demonstration on the [clj-refactor.el Wiki](https://github.com/clojure-emacs/clj-refactor.el/wiki/cljr-thread-last-all).
+
+* Fully unwind a threaded expression.
+
+Unwind and remove the threading macro. See demonstration on the [clj-refactor.el Wiki](https://github.com/clojure-emacs/clj-refactor.el/wiki/cljr-unwind-all).
 
 ## Related packages
 
--- a/clojure-mode.el
+++ b/clojure-mode.el
@@ -195,6 +195,11 @@ Out-of-the box clojure-mode understands lein, boot and gradle."
   (let ((map (make-sparse-keymap)))
     (define-key map (kbd "C-:") #'clojure-toggle-keyword-string)
     (define-key map (kbd "C-c SPC") #'clojure-align)
+    (define-key map (kbd "C-c C-r t") #'clojure-thread)
+    (define-key map (kbd "C-c C-r u") #'clojure-unwind)
+    (define-key map (kbd "C-c C-r f") #'clojure-thread-first-all)
+    (define-key map (kbd "C-c C-r l") #'clojure-thread-last-all)
+    (define-key map (kbd "C-c C-r a") #'clojure-unwind-all)
     (easy-menu-define clojure-mode-menu map "Clojure Mode Menu"
       '("Clojure"
         ["Toggle between string & keyword" clojure-toggle-keyword-string]
@@ -204,6 +209,13 @@ Out-of-the box clojure-mode understands lein, boot and gradle."
         ["Update ns form" clojure-update-ns]
         "--"
         ["Align expression" clojure-align]
+        "--"
+        ("Refactor -> and ->>"
+         ["Fully thread a form with ->" clojure-thread-first-all]
+         ["Fully thread a form with ->>" clojure-thread-last-all]
+         ["Fully unwind a threading macro" clojure-unwind-all]
+         ["Thread once more" clojure-thread]
+         ["Unwind once" clojure-unwind])
         "--"
         ["Version" clojure-mode-display-version]))
     map)
@@ -1533,6 +1545,178 @@ This will skip over sexps that don't represent objects, so that ^hints and
                         (clojure--looking-at-non-logical-sexp))))
           (backward-sexp 1))
         (setq n (1- n))))))
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;
+;; Refactoring support
+;;
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+(defcustom clojure-thread-all-but-last nil
+  "When true `cljr-thread-first-all' and `cljr-thread-last-all' don't thread the last expression."
+  :package-version '(clojure-mode . "5.4.0")
+  :safe #'booleanp
+  :type 'boolean)
+
+(defun clojure--unwind-last ()
+  (forward-sexp)
+  (save-excursion
+    (let ((contents (clojure-delete-and-extract-sexp)))
+      (when (looking-at " *\n")
+        (join-line -1))
+      (clojure--ensure-parens-around-function-names)
+      (let* ((sexp-beg-line (line-number-at-pos))
+             (sexp-end-line (progn (forward-sexp)
+                                   (line-number-at-pos)))
+             (multiline-sexp-p (not (= sexp-beg-line sexp-end-line))))
+        (down-list -1)
+        (when multiline-sexp-p
+          (newline))
+        (insert contents)
+        (when multiline-sexp-p
+          (clojure-indent-line)))))
+  (forward-char))
+
+(defun clojure--ensure-parens-around-function-names ()
+  (clojure--looking-at-non-logical-sexp)
+  (unless (looking-at "(")
+    (insert-parentheses 1)
+    (backward-up-list)))
+
+(defun clojure--unwind-first ()
+  "Unwind a thread first macro once.
+Point must be between the opening paren and the -> symbol."
+  (forward-sexp)
+  (save-excursion
+    (let ((contents (clojure-delete-and-extract-sexp)))
+      (when (looking-at " *\n")
+        (join-line -1))
+      (clojure--ensure-parens-around-function-names)
+      (down-list)
+      (forward-sexp)
+      (insert contents)))
+  (forward-char))
+
+(defun clojure--pop-out-of-threading ()
+  (save-excursion
+    (down-list 2)
+    (backward-up-list)
+    (raise-sexp)
+    (let ((beg (point))
+          (end (progn
+                 (forward-sexp)
+                 (point))))
+      (clojure-indent-region beg end))))
+
+(defun clojure--nothing-more-to-unwind ()
+  (save-excursion
+    (let ((beg (point)))
+      (forward-sexp)
+      (down-list -1)
+      (backward-sexp 2) ;; the last sexp, the threading macro
+      (when (looking-back "(\\s-*")
+          (backward-up-list)) ;; and the paren
+      (= beg (point)))))
+
+;;;###autoload
+(defun clojure-unwind ()
+  "Unwind thread at point or above point by one level.
+Return nil if there are no more levels to unwind."
+  (interactive)
+  (ignore-errors
+    (when (looking-at "(")
+      (forward-char 1)
+      (forward-sexp 1)))
+  (search-backward-regexp "([^-]*->")
+  (if (clojure--nothing-more-to-unwind)
+      (progn (clojure--pop-out-of-threading)
+             nil)
+    (down-list)
+    (cond
+     ((looking-at "[^-]*->\\_>")  (clojure--unwind-first))
+     ((looking-at "[^-]*->>\\_>") (clojure--unwind-last)))
+    t))
+
+;;;###autoload
+(defun clojure-unwind-all ()
+  "Fully unwind thread at point or above point."
+  (interactive)
+  (while (clojure-unwind)))
+
+(defun clojure--remove-superfluous-parens ()
+  (when (looking-at "([^ )]+)")
+    (delete-pair)))
+
+(defun clojure--thread-first ()
+  (down-list)
+  (forward-symbol 1)
+  (unless (looking-at ")")
+    (let ((contents (clojure-delete-and-extract-sexp)))
+      (backward-up-list)
+      (just-one-space 0)
+      (insert contents)
+      (newline-and-indent)
+      (clojure--remove-superfluous-parens)
+      t)))
+
+(defun clojure--thread-last ()
+  (forward-sexp 2)
+  (down-list -1)
+  (backward-sexp)
+  (unless (looking-back "(")
+    (let ((contents (clojure-delete-and-extract-sexp)))
+      (just-one-space 0)
+      (backward-up-list)
+      (insert contents)
+      (newline-and-indent)
+      (clojure--remove-superfluous-parens)
+      ;; cljr #255 Fix dangling parens
+      (backward-up-list)
+      (forward-sexp)
+      (when (looking-back "^\\s-*)+\\s-*")
+        (join-line))
+      t)))
+
+(defun clojure--threadable-p ()
+  (save-excursion
+    (forward-symbol 1)
+    (looking-at "[\n\r\t ]*(")))
+
+;;;###autoload
+(defun clojure-thread ()
+  "Thread by one more level an existing threading macro."
+  (interactive)
+  (ignore-errors
+    (when (looking-at "(")
+      (forward-char 1)
+      (forward-sexp 1)))
+  (search-backward-regexp "([^-]*->")
+  (down-list)
+  (when (clojure--threadable-p)
+    (cond
+     ((looking-at "[^-]*->\\_>")  (clojure--thread-first))
+     ((looking-at "[^-]*->>\\_>") (clojure--thread-last)))))
+
+(defun clojure--thread-all (first-or-last-thread but-last)
+  (save-excursion
+    (insert-parentheses 1)
+    (insert first-or-last-thread))
+  (while (save-excursion (clojure-thread)))
+  (when (or but-last clojure-thread-all-but-last)
+    (clojure-unwind)))
+
+;;;###autoload
+(defun clojure-thread-first-all (but-last)
+  "Fully thread the form at point using ->.
+When BUT-LAST is passed the last expression is not threaded."
+  (interactive "P")
+  (clojure--thread-all "-> " but-last))
+
+;;;###autoload
+(defun clojure-thread-last-all (but-last)
+  "Fully thread the form at point using ->>.
+When BUT-LAST is passed the last expression is not threaded."
+  (interactive "P")
+  (clojure--thread-all "->> " but-last))
 
 (defconst clojurescript-font-lock-keywords
   (eval-when-compile
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 * When aligning forms with `clojure-align` (or with the automatic align feature), blank lines will divide alignment regions.
 * [#378](https://github.com/clojure-emacs/clojure-mode/issues/378): Font-lock escape characters in strings.
+* Port threading macros related features from clj-refactor.el. Available refactorings: thread, unwind, thread first all, thread last all, unwind all.
 
 ## 5.3.0 (2016-04-04)
 
--- a/test/clojure-mode-refactor-threading-test.el
+++ b/test/clojure-mode-refactor-threading-test.el
@@ -1,0 +1,477 @@
+;;; clojure-mode-refactor-threading-test.el --- Clojure Mode: refactor threading tests  -*- lexical-binding: t; -*-
+
+;; Copyright (C) 2016 Benedek Fazekas <benedek.fazekas@gmail.com>
+
+;; This file is not part of GNU Emacs.
+
+;; This program is free software; you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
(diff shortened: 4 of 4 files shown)


README.md
```diff
@@ -220,6 +222,32 @@ This can also be done automatically (as part of indentation) by
...
+
+Supports both `->>` and `->`. See demonstration on the [clj-refactor.el Wiki](https://github.com/clojure-emacs/clj-refactor.el/wiki/cljr-unwind-thread).
+
+* Wrap in thread first (`->`) and fully thread.
+
+Introduce the thread first macro and rewrite the entire form. With a prefix argument do not thread the last form. See demonstration on the [clj-refactor.el Wiki](https://github.com/clojure-emacs/clj-refactor.el/wiki/cljr-thread-first-all).
+
+* Wrap in thread last (`->>`) and fully thread.
+
+Introduce the thread last amcro and rewrite the entire form. With a prefix argument do not thread the last form. See demonstration on the [clj-refactor.el Wiki](https://github.com/clojure-emacs/clj-refactor.el/wiki/cljr-thread-last-all).
```
bbatsov: amcro -> macro

clojure-mode.el
```diff
@@ -205,6 +210,12 @@ Out-of-the box clojure-mode understands lein, boot and gradle."
         "--"
         ["Align expression" clojure-align]
         "--"
+        ["Thread an existing threading macro once more" clojure-thread]
```
bbatsov: Guess this can be a separate submenu.

clojure-mode.el
```diff
@@ -195,6 +195,11 @@ Out-of-the box clojure-mode understands lein, boot and gradle."
   (let ((map (make-sparse-keymap)))
     (define-key map (kbd "C-:") #'clojure-toggle-keyword-string)
     (define-key map (kbd "C-c SPC") #'clojure-align)
+    (define-key map (kbd "C-c th") #'clojure-thread)
```
bbatsov: We should change the defaults to something more common - e.g. `C-c C-r` something.

bbatsov:
Nice! I'll go through the code in more details later today or tomorrow.

Malabarba:
Me too!

clojure-mode.el
benedekfazekas: I was selfish here as I have the `cljr` prefix set to `C-c` too. Also the two letter mnemonic works so why change it?
So what are you suggesting @bbatsov `C-c C-r th`? is not that a bit tooo long?

Malabarba: They should all be a single key under a 2-key predix. For instance, I would suggest the keys be `t`, `u`, `f`, `l`, and `a`. And the prefix could be `C-c C-r` or `C-c t` or something else even.

clojure-mode.el
```diff
@@ -1534,6 +1546,197 @@ This will skip over sexps that don't represent objects, so that ^hints and
...
+             (multiline-sexp-p (not (= sexp-beg-line sexp-end-line))))
+        (down-list -1)
+        (when multiline-sexp-p
+          (newline))
+        (insert contents)
+        (when multiline-sexp-p
+          (clojure-indent-line)))))
+  (forward-char))
+
+(defun clojure--wrap-around ()
```
Malabarba: Is this the same as `(insert-parentheses 1)`?

clojure-mode.el
```diff
@@ -1534,6 +1546,197 @@ This will skip over sexps that don't represent objects, so that ^hints and
...
+  (insert (format "(%s)" (clojure-delete-and-extract-sexp))))
+
+(defun clojure--ensure-parens-around-function-names ()
+  (unless (looking-at "[\n\r\t ]?(")
+    (skip-syntax-forward " ")
```
Malabarba: Maybe these two lines should be:

```elisp
(skip-syntax-forward "\n\r\t ")
(unless (looking-at "(")
  ...)
```

clojure-mode.el
```diff
@@ -1534,6 +1546,197 @@ This will skip over sexps that don't represent objects, so that ^hints and
...
+                 (point))))
+      (clojure-indent-region beg end))))
+
+(defun clojure--nothing-more-to-unwind ()
+  (save-excursion
+    (let ((beg (point)))
+      (forward-sexp)
+      (down-list -1)
+      (backward-sexp 2) ;; the last sexp, the threading macro
+      (when (looking-back "(\\s-*")
```
Malabarba: IIUC, this entire `(when ...)` clause can be replaced with `(forward-char -1)`.

clojure-mode.el
```diff
@@ -1534,6 +1546,197 @@ This will skip over sexps that don't represent objects, so that ^hints and
...
+          (backward-up-list)) ;; and the paren
+      (= beg (point)))))
+
+;;;###autoload
+(defun clojure-unwind ()
+  "Unwind thread at point or above point by one level.
+Return nil if there are no more levels to unwind."
+  (interactive)
+  (ignore-errors
+    (forward-char 3))
```
Malabarba: `forward-char 3` might not work as expected for longer macros like `some->>`.
I would rather do 
```
(when (looking-at "(")
  (forward-char 1)
  (forward-sexp 1))
```

clojure-mode.el
```diff
@@ -1534,6 +1546,197 @@ This will skip over sexps that don't represent objects, so that ^hints and
...
+     ((looking-at "[^-]*->[\n\r\t ]")  (clojure--unwind-first))
+     ((looking-at "[^-]*->>[\n\r\t ]") (clojure--unwind-last)))
+    t))
+
+;;;###autoload
+(defun clojure-unwind-all ()
+  "Fully unwind thread at point or above point."
+  (interactive)
+  (while (clojure-unwind)
+    t))
```
Malabarba: This `t` is unnecessary.

clojure-mode.el
```diff
@@ -1534,6 +1546,197 @@ This will skip over sexps that don't represent objects, so that ^hints and
...
+
+(defun clojure--remove-superfluous-parens ()
+  (when (looking-at "([^ )]+)")
+    (down-list)
+    (raise-sexp)))
```
Malabarba: The two lines above are equivalent to `delete-pair`

clojure-mode.el
```diff
@@ -1534,6 +1546,197 @@ This will skip over sexps that don't represent objects, so that ^hints and
...
+        (join-line -1))
+      (clojure--ensure-parens-around-function-names)
+      (down-list)
+      (forward-sexp)
+      (insert contents)))
+  (forward-char))
+
+(defun clojure--pop-out-of-threading ()
+  (down-list 2)
+  (backward-up-list)
```
Malabarba: Going down-list twice and then back up is a little odd and will fail in some corner cases.
I think `(down-list 1) (forward-sexp 1)` would be safer.

clojure-mode.el
```diff
@@ -1534,6 +1546,197 @@ This will skip over sexps that don't represent objects, so that ^hints and
...
+      (clojure-indent-region beg end))))
+
+(defun clojure--nothing-more-to-unwind ()
+  (save-excursion
+    (let ((beg (point)))
+      (forward-sexp)
+      (down-list -1)
+      (backward-sexp 2) ;; the last sexp, the threading macro
+      (when (looking-back "(\\s-*")
+          (backward-up-list)) ;; and the paren
```
Malabarba: IIUC, this entire `(when ...)` clause can be replaced with `(forward-char -1)`.

clojure-mode.el
```diff
@@ -1534,6 +1546,197 @@ This will skip over sexps that don't represent objects, so that ^hints and
...
+      (down-list)
+      (forward-sexp)
+      (insert contents)))
+  (forward-char))
+
+(defun clojure--pop-out-of-threading ()
+  (down-list 2)
+  (backward-up-list)
+  (raise-sexp)
+  (save-excursion
```
Malabarba: This `save-excursion` should probably be surrounding the whole function.

clojure-mode.el
```diff
@@ -1534,6 +1546,197 @@ This will skip over sexps that don't represent objects, so that ^hints and
...
         (setq n (1- n))))))
 
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;
+;; Refactoring support
+;;
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+(defcustom clojure-thread-all-but-last nil
+  "When true `cljr-thread-first-all' and `cljr-thread-last-all' don't thread the last expression."
+  :package-version '(clojure-mode . "5.3.0")
```
Malabarba: 5.3 is already out. @bbatsov what's the next version?

clojure-mode.el
```diff
@@ -205,6 +210,13 @@ Out-of-the box clojure-mode understands lein, boot and gradle."
         "--"
         ["Align expression" clojure-align]
         "--"
+        ("Threading macro refactorings"
+         ["Thread an existing threading macro once more" clojure-thread]
+         ["Unwind a threading macro once" clojure-unwind]
+         ["Fully thread the form at point using ->" clojure-thread-first-all]
+         ["Fully thread the form at point using ->>" clojure-thread-last-all]
```
Malabarba: These two items should be the first two in the menu. And I think they should say simply:

> Thread a form with ->
> Thread a form with ->>

clojure-mode.el
```diff
@@ -205,6 +210,13 @@ Out-of-the box clojure-mode understands lein, boot and gradle."
         "--"
         ["Align expression" clojure-align]
         "--"
+        ("Threading macro refactorings"
+         ["Thread an existing threading macro once more" clojure-thread]
```
Malabarba: > Thread a form once more

clojure-mode.el
```diff
@@ -205,6 +210,13 @@ Out-of-the box clojure-mode understands lein, boot and gradle."
         "--"
         ["Align expression" clojure-align]
         "--"
+        ("Threading macro refactorings"
```
Malabarba: I prefer to call this `-> and ->> refactoring`

clojure-mode.el
```diff
@@ -1534,6 +1546,197 @@ This will skip over sexps that don't represent objects, so that ^hints and
...
+  (interactive)
+  (ignore-errors
+    (forward-char 3))
+  (search-backward-regexp "([^-]*->")
+  (if (clojure--nothing-more-to-unwind)
+      (progn (clojure--pop-out-of-threading)
+             nil)
+    (down-list)
+    (cond
+     ((looking-at "[^-]*->[\n\r\t ]")  (clojure--unwind-first))
```
Malabarba: `"\\_<[^ ]*->\\_>"`

clojure-mode.el
```diff
@@ -1534,6 +1546,197 @@ This will skip over sexps that don't represent objects, so that ^hints and
...
+     ((looking-at "[^-]*->>[\n\r\t ]") (clojure--unwind-last)))
```
Malabarba: same here

clojure-mode.el
```diff
@@ -1534,6 +1546,197 @@ This will skip over sexps that don't represent objects, so that ^hints and
...
+(defun clojure--wrap-around ()
+  (insert (format "(%s)" (clojure-delete-and-extract-sexp))))
+
+(defun clojure--ensure-parens-around-function-names ()
+  (unless (looking-at "[\n\r\t ]?(")
+    (skip-syntax-forward " ")
+    (clojure--wrap-around)
+    (backward-sexp)))
+
+(defun clojure--unwind-first ()
```
Malabarba: A short docstring would be super helpful here. Just something that says:

> Unwind a -> once.
> Point must be between the opening paren and the -> symbol.

clojure-mode.el
```diff
@@ -1534,6 +1546,197 @@ This will skip over sexps that don't represent objects, so that ^hints and
...
+     ((looking-at "[^-]*->>[\n\r\t ]") (clojure--thread-last)))))
+
+(defun clojure--thread-all (first-or-last-thread but-last)
+  (save-excursion
+    (clojure--wrap-around)
+    (backward-list)
+    (down-list)
+    (insert first-or-last-thread))
+  (while (save-excursion (clojure-thread))
+    t)
```
Malabarba: Again, irrelevant.

clojure-mode.el
Malabarba: Actually, calling `(clojure--looking-at-non-logical-sexp)` would skip both whitespace *and* comments, so it might be a better replacement for `(skip-syntax-forward "\n\r\t ")`.

clojure-mode.el
```diff
@@ -1534,6 +1546,197 @@ This will skip over sexps that don't represent objects, so that ^hints and
...
+
+;;;###autoload
+(defun clojure-thread ()
+  "Thread by one more level an existing threading macro."
+  (interactive)
+  (when (looking-at "(?[^-]*-?>")
+    (goto-char (match-end 0)))
+  (search-backward-regexp "([^-]*->")
+  (down-list)
+  (if (not (clojure--thread-guard))
```
Malabarba: `(when (clojure--thread-guard))`

clojure-mode.el
```diff
@@ -1534,6 +1546,197 @@ This will skip over sexps that don't represent objects, so that ^hints and
...
+    (insert first-or-last-thread))
+  (while (save-excursion (clojure-thread))
+    t)
+  (when (or but-last clojure-thread-all-but-last)
+    (clojure-unwind)))
+
+;;;###autoload
+(defun clojure-thread-first-all (but-last)
+  "Fully thread the form at point using ->.
+
```
Malabarba: No need for a blank line in a 2-line docstring.

clojure-mode.el
```diff
@@ -1534,6 +1546,197 @@ This will skip over sexps that don't represent objects, so that ^hints and
...
+When BUT-LAST is passed the last expression is not threaded."
+  (interactive "P")
+  (clojure--thread-all "-> " but-last))
+
+;;;###autoload
+(defun clojure-thread-last-all (but-last)
+  "Fully thread the form at point using ->>.
+
```
Malabarba: same here

clojure-mode.el
```diff
@@ -1534,6 +1546,197 @@ This will skip over sexps that don't represent objects, so that ^hints and
...
+    (if (looking-at "[\n\r\t ]*(")
+        t
+      nil)))
+
+;;;###autoload
+(defun clojure-thread ()
+  "Thread by one more level an existing threading macro."
+  (interactive)
+  (when (looking-at "(?[^-]*-?>")
+    (goto-char (match-end 0)))
```
Malabarba: On line 1627 you use a different approach to do the same thing. Decide which one's best and use the same on both sides.

clojure-mode.el
```diff
@@ -1534,6 +1546,197 @@ This will skip over sexps that don't represent objects, so that ^hints and
...
+
+(defun clojure--thread-last ()
+  (forward-sexp 2)
+  (down-list -1)
+  (let* ((end (point))
+         (beg (progn (backward-sexp)
+                     (point)))
+         (contents (buffer-substring beg end)))
+    (if (looking-back "(")
+        nil
```
Malabarba: `(unless (looking-back "(")`

clojure-mode.el
```diff
@@ -1534,6 +1546,197 @@ This will skip over sexps that don't represent objects, so that ^hints and
...
+      (when (looking-back "^\\s-*)+\\s-*")
+        (join-line))
+      t)))
+
+(defun clojure--thread-guard ()
+  (save-excursion
+    (forward-symbol 1)
+    (if (looking-at "[\n\r\t ]*(")
+        t
+      nil)))
```
Malabarba: This entire conditional is identical to `(looking-at "[\n\r\t ]*(")`

clojure-mode.el
```diff
@@ -1534,6 +1546,197 @@ This will skip over sexps that don't represent objects, so that ^hints and
...
+      (newline-and-indent)
+      (clojure--remove-superfluous-parens)
+      ;; cljr #255 Fix dangling parens
+      (backward-up-list)
+      (forward-sexp)
+      (when (looking-back "^\\s-*)+\\s-*")
+        (join-line))
+      t)))
+
+(defun clojure--thread-guard ()
```
Malabarba: This either needs a docstring or a better name. I have no idea what's its purpose.

clojure-mode.el
```diff
@@ -1534,6 +1546,197 @@ This will skip over sexps that don't represent objects, so that ^hints and
...
+;;;###autoload
+(defun clojure-thread ()
+  "Thread by one more level an existing threading macro."
+  (interactive)
+  (when (looking-at "(?[^-]*-?>")
+    (goto-char (match-end 0)))
+  (search-backward-regexp "([^-]*->")
+  (down-list)
+  (if (not (clojure--thread-guard))
+      nil
```
Malabarba: `(when (clojure--thread-guard))`

clojure-mode.el
```diff
@@ -1534,6 +1546,197 @@ This will skip over sexps that don't represent objects, so that ^hints and
...
+    (cond
+     ((looking-at "[^-]*->[\n\r\t ]")  (clojure--thread-first))
```
Malabarba: `"\\_<[^ ]*->\\_>"`

clojure-mode.el
```diff
@@ -1534,6 +1546,197 @@ This will skip over sexps that don't represent objects, so that ^hints and
...
+     ((looking-at "[^-]*->>[\n\r\t ]") (clojure--thread-last)))))
```
Malabarba: same here

clojure-mode.el
```diff
@@ -1534,6 +1546,197 @@ This will skip over sexps that don't represent objects, so that ^hints and
...
+        t
+      nil)))
+
+;;;###autoload
+(defun clojure-thread ()
+  "Thread by one more level an existing threading macro."
+  (interactive)
+  (when (looking-at "(?[^-]*-?>")
+    (goto-char (match-end 0)))
+  (search-backward-regexp "([^-]*->")
```
Malabarba: `([^ ]*->>?\\_>`

clojure-mode.el
Malabarba: I think I understand now. A good name might be `clojure--threadable-form-p`.

clojure-mode.el
```diff
@@ -1534,6 +1546,197 @@ This will skip over sexps that don't represent objects, so that ^hints and
...
+      (just-one-space 0)
+      (backward-up-list)
+      (insert contents)
+      (newline-and-indent)
+      (clojure--remove-superfluous-parens)
+      ;; cljr #255 Fix dangling parens
+      (backward-up-list)
+      (forward-sexp)
+      (when (looking-back "^\\s-*)+\\s-*")
+        (join-line))
```
Malabarba: Not sure if that can happen, but this will break the code if the line above ends in a comment.

clojure-mode.el
```diff
@@ -1534,6 +1546,197 @@ This will skip over sexps that don't represent objects, so that ^hints and
...
+      (clojure--remove-superfluous-parens)
+      t)))
+
+(defun clojure--thread-last ()
+  (forward-sexp 2)
+  (down-list -1)
+  (let* ((end (point))
+         (beg (progn (backward-sexp)
+                     (point)))
+         (contents (buffer-substring beg end)))
```
Malabarba: I think you can remove all of these let-bindings. Instead, you can use a single let-bind inside the conditional by calling `clojure-delete-and-extract-sexp`.

clojure-mode.el
```diff
@@ -1534,6 +1546,197 @@ This will skip over sexps that don't represent objects, so that ^hints and
...
+    (if (string= contents ")")
+        nil
+      (delete-region beg end)
+      (backward-up-list)
+      (just-one-space 0)
+      (insert contents)
+      (newline-and-indent)
+      (clojure--remove-superfluous-parens)
+      t)))
```
Malabarba:
```elisp
(defun clojure--thread-first ()
  (down-list)
  (forward-symbol 1)
  (unless (looking-at ")")
    (let ((contents (clojure-delete-and-extract-sexp)))
      (backward-up-list)
      (just-one-space 0)
      (insert contents)
      (newline-and-indent)
      (clojure--remove-superfluous-parens)
      t)))
```

Malabarba:
Ok. I believe those are all the comments I have. Sorry to give you so much work to do. :sweat_smile:

clojure-mode.el
benedekfazekas: `(forward-char -1)` would not do the right thing if there is whitespace between the `(` and the threading macro I think.

clojure-mode.el
benedekfazekas: not sure about this one. we are the beginning of a list: `(|-> foobar)`. your pattern does not match if there is whitespace between the paren and the threading macro...

clojure-mode.el
benedekfazekas: as above

clojure-mode.el
benedekfazekas: as for line 1634

clojure-mode.el
benedekfazekas: as above

clojure-mode.el
benedekfazekas: again

benedekfazekas:
thanks @Malabarba for the thorough review. Mostly I went with your recommendations see my comments, questions etc in comments. Ta!

clojure-mode.el
magnars: `(down-list 1) (forward-sexp 1)` is not equivalent to `(down-list 2) (backward-up-list)`. There might be more than one sexp in front of the next pair.

clojure-mode.el
Malabarba: I see. Then maybe `" *\\_<[^ ]*->\\_>"`? Or just `"[^-]*->\\_>"`. My point was just to use `\\_>` instead of `[\n\r\t ]`.

clojure-mode.el
Malabarba: Good point. :+1:

clojure-mode.el
benedekfazekas: Although I am not sure that is right @magnars as when `clojure--pop-out-of-threading` is called there is nothing more to thread so the only thing should be the `->` or the `->>` I still reverted this change because don't want to build this assumption into the impl of this function.

benedekfazekas:
afaik all comments are dealt with except the one asking @bbatsov about the next clojure-mode version.

clojure-mode.el
bbatsov: 5.4

bbatsov:
5.4.0

Malabarba:
Great! Thanks Benedek!

Malabarba:
Just realized that `C-c C-r` was a bad choice on our part. That's bound to `eval-region` in cider (though, honestly, maybe we should change the later as it's not important enough to deserve a 2-key combo).

bbatsov:
I think we should simply convert `C-c C-e` to a prefix map and stuff there all evaluation keybindings. We'll leave just `C-c C-c`, `C-M-x` and `C-x C-e` as shorter eval keybindings due to their popularity.

Malabarba:
I have a feeling a lot of people use the C-c C-e binding. It's easy, useful, and makes more sense than C-x C-e to anyone who doesn't program elisp.

bbatsov:
> I have a feeling a lot of people use the C-c C-e binding. It's easy, useful, and makes more sense than C-x C-e to anyone who doesn't program elisp.

I never use it, so from my POV dropping it seems reasonable. In general I've always been against having multiple keybindings for the same action. Half the eval commands are hardly ever being used - freeing up some keybindings would be pretty great.

Malabarba:
I know. My muscle memory always uses C-x C-e too, but you and I are Emacsers first. :-)
I think C-c C-e is more natural for those who are Clojurians first, because it's under C-c like every other one of our keybinds.

bbatsov:
Maybe. But we can make it `C-c C-e e` which is not that bad all things considered. Then we'll also have `C-c C-e d` (defun), `C-c C-e b` (buffer), `C-c C-e n` (namespace), `C-c C-e r` (region), etc. To some extent this sounds sensible, but I have little doubt some people will be unhappy. :-)

Malabarba:
> To some extent this sounds sensible, but I have little doubt some people will be unhappy. :-)

Exactly. :disappointed: I just wish there was another prefix that made sense here.

bbatsov:
Maybe we can repurpose `C-c C-v` for this keymap.

Malabarba:
I think I would prefer that. `C-c C-v` is a very new command, so it's less likely that people will be attached to it. :-)

bbatsov:
My thought exactly. :-)

PR #381, opened 2016-05-26
Add cycling privacy, collection type, if/if-not

benedekfazekas:
- [x] The commits are consistent with our [contribution guidelines](../CONTRIBUTING.md)

Migrate cycle privacy, cycle collection type and cycle if/if-not implementations from clj-refactor.el.

Additionally refactor `def-threading-test` macro to use it for testing cycling stuff too and fix duplicate test names in `clojure-mode-refactor-threading-test`

### DIFF
--- a/clojure-mode.el
+++ b/clojure-mode.el
@@ -205,6 +205,20 @@ Out-of-the box clojure-mode understands lein, boot and gradle."
     (define-key map (kbd "C-c C-r l") #'clojure-thread-last-all)
     (define-key map (kbd "C-c C-r C-a") #'clojure-unwind-all)
     (define-key map (kbd "C-c C-r a") #'clojure-unwind-all)
+    (define-key map (kbd "C-c C-r C-p") #'clojure-cycle-privacy)
+    (define-key map (kbd "C-c C-r p") #'clojure-cycle-privacy)
+    (define-key map (kbd "C-c C-r C-(") #'clojure-convert-collection-to-list)
+    (define-key map (kbd "C-c C-r (") #'clojure-convert-collection-to-list)
+    (define-key map (kbd "C-c C-r C-'") #'clojure-convert-collection-to-quoted-list)
+    (define-key map (kbd "C-c C-r '") #'clojure-convert-collection-to-quoted-list)
+    (define-key map (kbd "C-c C-r C-{") #'clojure-convert-collection-to-map)
+    (define-key map (kbd "C-c C-r {") #'clojure-convert-collection-to-map)
+    (define-key map (kbd "C-c C-r C-[") #'clojure-convert-collection-to-vector)
+    (define-key map (kbd "C-c C-r [") #'clojure-convert-collection-to-vector)
+    (define-key map (kbd "C-c C-r C-#") #'clojure-convert-collection-to-set)
+    (define-key map (kbd "C-c C-r #") #'clojure-convert-collection-to-set)
+    (define-key map (kbd "C-c C-r C-i") #'clojure-cycle-if)
+    (define-key map (kbd "C-c C-r i") #'clojure-cycle-if)
     (define-key map (kbd "C-c C-r n i") #'clojure-insert-ns-form)
     (define-key map (kbd "C-c C-r n h") #'clojure-insert-ns-form-at-point)
     (define-key map (kbd "C-c C-r n u") #'clojure-update-ns)
@@ -213,11 +227,19 @@ Out-of-the box clojure-mode understands lein, boot and gradle."
       '("Clojure"
         ["Toggle between string & keyword" clojure-toggle-keyword-string]
         ["Align expression" clojure-align]
+        ["Cycle privacy" clojure-cycle-privacy]
+        ["Cycle if, if-not" clojure-cycle-if]
         ("ns forms"
          ["Insert ns form at the top" clojure-insert-ns-form]
          ["Insert ns form here" clojure-insert-ns-form-at-point]
          ["Update ns form" clojure-update-ns]
          ["Sort ns form" clojure-sort-ns])
+        ("Convert collection"
+         ["Convert to list" clojure-convert-collection-to-list]
+         ["Convert to quoted list" clojure-convert-collection-to-quoted-list]
+         ["Convert to map" clojure-convert-collection-to-map]
+         ["Convert to vector" clojure-convert-collection-to-vector]
+         ["Convert to set" clojure-convert-collection-to-set])
         ("Refactor -> and ->>"
          ["Thread once more" clojure-thread]
          ["Fully thread a form with ->" clojure-thread-first-all]
@@ -1629,6 +1651,8 @@ This will skip over sexps that don't represent objects, so that ^hints and
 ;; Refactoring support
 ;;
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+;;; Threading macros related
 (defcustom clojure-thread-all-but-last nil
   "Non-nil means do not thread the last expression.
 This means that `clojure-thread-first-all' and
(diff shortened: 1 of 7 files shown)


clojure-mode.el
```diff
@@ -1824,6 +1836,114 @@ When BUT-LAST is passed the last expression is not threaded."
...
+;;;###autoload
+(defun clojure-cycle-collection ()
+  "Convert the coll at (point) from (x) -> {x} -> [x] -> -> #{x} -> (x) recur.
+
+See: https://github.com/clojure-emacs/clj-refactor.el/wiki/cljr-cycle-coll"
+  (interactive)
+  (save-excursion
+    (while (and
+            (> (point) 1)
+            (not (eq (string-to-char "(") (char-after)))
```
Malabarba: `?\(`

Malabarba: Actually. I'd prefer to replace this line and the 3 below with a single `looking-at`.

clojure-mode.el
```diff
@@ -1824,6 +1836,114 @@ When BUT-LAST is passed the last expression is not threaded."
...
+
+;;;###autoload
+(defun clojure-cycle-collection ()
+  "Convert the coll at (point) from (x) -> {x} -> [x] -> -> #{x} -> (x) recur.
+
+See: https://github.com/clojure-emacs/clj-refactor.el/wiki/cljr-cycle-coll"
+  (interactive)
+  (save-excursion
+    (while (and
+            (> (point) 1)
```
Malabarba: Replace 1 with point-min.

clojure-mode.el
```diff
@@ -1824,6 +1836,114 @@ When BUT-LAST is passed the last expression is not threaded."
...
+        (insert "(" (substring (clojure-delete-and-extract-sexp) 1 -1) ")")))
+
+     ((eq (string-to-char "[") (char-after))
+      (insert "#{" (substring (clojure-delete-and-extract-sexp) 1 -1) "}"))
+
+     ((equal 1 (point))
+      (user-error "Beginning of file reached, collection is not found")))))
+
+(defun clojure--goto-if ()
+  (when (in-string-p)
```
Malabarba: I'm relatively sure we don't have this function. :-)

clojure-mode.el
Malabarba: Actually, just replace the whole line with bobp.

clojure-mode.el
```diff
@@ -1824,6 +1836,114 @@ When BUT-LAST is passed the last expression is not threaded."
...
+    (while (and
+            (> (point) 1)
+            (not (eq (string-to-char "(") (char-after)))
+            (not (string= "#{" (buffer-substring (point) (+ 2 (point)))))
+            (not (eq (string-to-char "{") (char-after)))
+            (not (eq (string-to-char "[") (char-after))))
+      (backward-char))
+
+    (cond
+     ((eq (string-to-char "(") (char-after))
```
Malabarba: `(string-to-char "(") ` == `?\(`

clojure-mode.el
```diff
@@ -1824,6 +1836,114 @@ When BUT-LAST is passed the last expression is not threaded."
...
+      (if (not (equal (string-to-char "#") (char-before)))
+          (insert "[" (substring (clojure-delete-and-extract-sexp) 1 -1) "]")
+        (backward-char)
+        (delete-char 1)
+        (insert "(" (substring (clojure-delete-and-extract-sexp) 1 -1) ")")))
+
+     ((eq (string-to-char "[") (char-after))
+      (insert "#{" (substring (clojure-delete-and-extract-sexp) 1 -1) "}"))
+
+     ((equal 1 (point))
```
Malabarba: (bobp)

clojure-mode.el
```diff
@@ -1824,6 +1836,114 @@ When BUT-LAST is passed the last expression is not threaded."
...
+      (user-error "Beginning of file reached, collection is not found")))))
+
+(defun clojure--goto-if ()
+  (when (in-string-p)
+    (while (or (not (looking-at "("))
+               (in-string-p))
+      (backward-char)))
+  (while (not (looking-at "\\((if \\)\\|\\((if-not \\)"))
```
Malabarba: FTR, I like this method of going backward-up-list until we find what we're looking for (instead of doing search-backward). And I think this is what we should be using on all the other commands too (including the thread/unwind commands). But that's something for another time.

clojure-mode.el
```diff
@@ -1824,6 +1836,114 @@ When BUT-LAST is passed the last expression is not threaded."
...
+     ((and (not clojure-use-metadata-for-privacy)
+           (looking-at "(defn"))
+      (forward-char 5)
+      (insert "-"))
+     ((looking-at "(def ^:private")
+      (forward-char 5)
+      (delete-char 10))
+     ((looking-at "(def ")
+      (forward-char 5)
+      (insert "^:private ")))))
```
Malabarba: I would prefer something like this (if it passes the tests):

```el
(defun clojure-cycle-privacy ()
  "Make public the current private def, or vice-versa.
See: https://github.com/clojure-emacs/clj-refactor.el/wiki/cljr-cycle-privacy"
  (interactive)
  (save-excursion
    (ignore-errors (forward-char 7))
    (search-backward-regexp "(defn?\\(-\\| ^:private\\)?\\_>")
    (if (match-string 1)
        (replace-match "" nil nil nil 1)
      (goto-char (match-end 0))
      (insert (if (or clojure-use-metadata-for-privacy
                      (equal (match-string 0 "(def")))
                  " ^:private"
                "-")))))
```

Malabarba:
Besides the comments. LGTM.

expez:
Btw, I recently used this feature in cljr and it occurred to me that we should remove the cycle to and from lists.  

Lists aren't used for literal data in clojure, vectors are, so it makes no sense to cycle to and from this collection type imo.  

Furthermore, at least in cljr, the list isn't quoted so that part of the refactoring is broken.  It's cycling between code and data now :/

tanzoniteblack:
@expez I rather like having lists in there, as I use it to cycle collections to to function calls quite often. So we're changing between code and data...but that's actually pretty useful, since code is data in clojure.

benedekfazekas:
thanks for the feedback @Malabarba @expez @tanzoniteblack 

re. cycle collection type I think both @expez and @tanzoniteblack have a point here. I don't see cycling between code and data a big problem either, in certain contexts (reagent for example) even stuff in a vector can be an fn invocation.

so I propose to actually *add* `'()` to the list of collection types and also leave `()` there. let me know what you think.

clojure-mode.el
benedekfazekas: hm... deffo looks nicer. let me test it/work on it a bit more tho

expez:
> so I propose to actually add '() to the list of collection types and also leave () there. let me know what you think.

When you have to click <kbd>C-c C-m cc</kbd> 4 times to get the collection you want you're not saving any time compared to a regular edit.  You're now proposing we add another one to cycle between, set, map, list, quoted list and vector.  I think this will be too much.

I'd even like to narrow it down even more, so it only cycles between vectors and sets.  Cycling to a map (from a vector/set) or from a map (to a vector/set) is something I've never done either.  Have any of you?

A refactoring isn't supposed to change the semantics of the code, but going from data to code certainly does.  I still think the cycle to `()` is rare enough that we can provide it as a separate function (with our without a keybinding) or even drop it completely.

benedekfazekas:
too much typing is a valid concern. so I can rewrite this not to be a `cycling` thing but rather you type `C-c C-r [` for a vector or `C-c C-r {` for a map. for set and quoted list we can use `C-c C-r #` and `C-c C-r '` respectively.

>  Cycling to a map (from a vector/set) or from a map (to a vector/set) is something I've never done either. Have any of you?

Unfortunately we don't have a way knowing how this (or any) feature is used out in the wild (special thanks @tanzoniteblack for your feedback!). I do remember cycling at least once to a map from a vector when I realised I wanted to use a different data structure. I have edited the map further after changing it to map from vector but this does not really matter I guess.

> A refactoring isn't supposed to change the semantics of the code

strictly speaking this is true. However, we do violate this anyway elsewhere, for example cycling privacy changes the public API of a namespaces which is not really a refactoring either strictly speaking... It is still a useful feature.

expez:
I really like you suggestion of using additional keybindings.  Those keybindings would be weird for anything else, but are easy to remember in this context. :+1:

benedekfazekas:
done from my side i think. although cycle collection is pretty much redone so comments welcome about it.

Malabarba:
I like the idea of switching to a specific collection. @bbatsov what do you think?

clojure-mode.el
```diff
@@ -1824,6 +1848,109 @@ When BUT-LAST is passed the last expression is not threaded."
...
+      (backward-char))
+
+    (when (or (eq ?\# (char-before))
+              (eq ?\' (char-before)))
+      (delete-char -1))
+
+    (when (and (bobp)
+               (not (memq (char-after) '(?\{ ?\( ?\[))))
+      (user-error "Beginning of file reached, collection is not found"))
+
```
Malabarba: Remove this blank line, and the two above.

clojure-mode.el
```diff
@@ -205,6 +205,20 @@ Out-of-the box clojure-mode understands lein, boot and gradle."
     (define-key map (kbd "C-c C-r l") #'clojure-thread-last-all)
     (define-key map (kbd "C-c C-r C-a") #'clojure-unwind-all)
     (define-key map (kbd "C-c C-r a") #'clojure-unwind-all)
+    (define-key map (kbd "C-c C-r C-p") #'clojure-cycle-privacy)
+    (define-key map (kbd "C-c C-r p") #'clojure-cycle-privacy)
+    (define-key map (kbd "C-c C-r (") #'clojure-convert-collection-to-list)
+    (define-key map (kbd "C-c C-r C-(") #'clojure-convert-collection-to-list)
```
Malabarba: A little silly, but could you write the `C-(` keybind above the `(` keybind? Same for the other collections.

Malabarba:
Besides the two new comments,  LGTM.

benedekfazekas:
no rush here guys, but I am not aware of anything left to do here...

Malabarba:
Thanks again Benedek!

PR #384, opened 2016-06-15
Fix bug when passing lambda to define-clojure-indent, closes #383

### LINKED ISSUE #383: Cannot pass lambdas to clojure-indent-function
## Expected behavior
The following should work in `define-clojure-indent`

`(trace-handlers (lambda (&rest _) 0))`

## Actual behavior
It fails to trigger the right case branch.

```
(define-clojure-indent
  (trace-handlers (lambda (&rest _) 0)))
```

```
clojure-mode (version 5.3.0)
```

The problem seems [here](https://github.com/clojure-emacs/clojure-mode/blob/master/clojure-mode.el#L1238) where `functionp` returns `nil` as `lambda` is a special form.

arichiardi:
@Malabarba I added a test and it passes :+1: 
Also, I tried against my codebase with success!

Thanks a lot!

### DIFF
--- a/test/clojure-mode-indentation-test.el
+++ b/test/clojure-mode-indentation-test.el
@@ -164,6 +164,17 @@ values of customisable variables."
 (ala/bala top
   |one)")
 
+;; we can pass a lambda to explicitely set the column
+(put-clojure-indent 'arsymbol (lambda (indent-point state) 0))
+
+(check-indentation symbol-with-lambda
+  "
+(arsymbol
+ |one)"
+  "
+(arsymbol
+|one)")
+
 (check-indentation form-with-metadata
   "
 (ns ^:doc app.core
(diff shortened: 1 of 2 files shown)


test/clojure-mode-indentation-test.el
```diff
@@ -164,8 +164,19 @@ values of customisable variables."
...
+(check-indentation symbol-with-lambda
+  "
+(arsymbol
+ |one)"
   "
+(arsymbol
+|one)")
+
+(check-indentation form-with-metadata
+                   "
```
Malabarba: Indent looks off here.

arichiardi: Ok well, it seems aggressive mode was not disabled after all

arichiardi:
It should be ok now, thanks for double checking :wink:

Malabarba:
Great! Thanks for the tests.

PR #385, opened 2016-06-17
Add initial support for smartparens-mode.

hrathod:
I have added some functionality to make `smartparens-mode` play nicely with `clojure-mode` by making the literal set notation and lambda shorthand notation easier to work with.

### DIFF
--- a/clojure-mode.el
+++ b/clojure-mode.el
@@ -342,6 +342,13 @@ instead of to `clojure-mode-map'."
     (add-to-list 'paredit-space-for-delimiter-predicates
                  #'clojure-no-space-after-tag)))
 
+(defun clojure-smartparens-setup ()
+  "Make \"smartparens-mode\" play nicely with `clojure-mode'."
+  (when (and (fboundp 'sp-with-modes) (fboundp 'sp-local-pair))
+    (sp-with-modes '(clojure-mode)
+      (sp-local-pair "#{" "}")
+      (sp-local-pair "#(" ")"))))
+
 (defun clojure-mode-variables ()
   "Set up initial buffer-local variables for Clojure mode."
   (setq-local imenu-create-index-function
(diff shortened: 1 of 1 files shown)


bbatsov:
This was merged before I could comment, but isn't setup like this one usually bundled with `smartparens` itself?

expez:
> This was merged before I could comment, but isn't setup like this one usually bundled with smartparens itself?

Yes.  It doesn't belong here.

Fuco1:
We don't have setup for clojure yet though.  @expez would you like to port this over?

Malabarba:
Sorry. Since we had something similar for paredit, I thought it was straightforward to do the same for smartparens.

bbatsov:
The only reason we have the paredit code is that it's pretty hard to push any changes to paredit itself. I'm not opposed to having the smartparens code in clojure-mode, but I do believe it's best to follow the common sp conventions and the bundle the necessary config with it.

Malabarba:
👍 I agree.

hrathod:
Shall I create a new PR to revert these changes, then?  Or has that already been done?

Malabarba:
I can revert the changes here. I was just waiting for someone to open a PR on https://github.com/Fuco1/smartparens

Malabarba:
> I can revert the changes here. I was just waiting for someone to open a PR on https://github.com/Fuco1/smartparens

Since it is apparently agreed upon that this code should go in smartparens, and since this is causing other issues (#386) I decided not to wait.

hrathod:
Sorry for the troubles!

Malabarba:
Not at all! Thanks for the effort. :-)
The troubles were my fault.

Fuco1:
In the meantime @vspinu has started (probably independently) an effort to add some settings: https://github.com/Fuco1/smartparens/issues/593

Anyone want to jump in?

PR #387, opened 2016-06-20
Revert "Add initial support for smartparens-mode."

### LINKED ISSUE #386: New smartparens support breaks Spacemacs
Downstream bug report is here: syl20bnr/spacemacs#6375

Malabarba:
Reverts clojure-emacs/clojure-mode#385
Fixes #386


Malabarba:
The reasons for reverting are discussed in #385

PR #391, opened 2016-06-29
Symbols in Clojure are allowed to contain `!' and `?'

### LINKED ISSUE #392: wrong syntax class for `?` and `!`
## Expected behavior 
When pressing `C-c C-d d` while point is inside `swap!` I expect to be prompted for the docs about `swap!`

## Actual behavior
I get prompted for the docs about `swap`

Press `C-c C-d d` and be prompted for docs about `swap`

Same goes for symbols ending with `?`

Executing the following elisp fixes the problems, but I don't know how much breakage this will incur.
```
(modify-syntax-entry ?! "w")
(modify-syntax-entry ?? "w")
```
### CIDER version information

```
;; CIDER 0.13.0snapshot (package: 20160524.714), nREPL 0.2.12
;; Clojure 1.8.0, Java 1.8.0_92
```

### Emacs version

 24.5.1

### Operating system

OS X 10.11 "El Capitan"

slipset:

- fixes #392 `cider-symbol-at-point` for functions containing `!` and `?`, like `swap!` and `nil?`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Bugs fixed
 
+* [#392](https://github.com/clojure-emacs/clojure-mode/issues/392): Wrong syntax class for `?` and `!`
+
 * [#389](https://github.com/clojure-emacs/clojure-mode/issues/389): Fixed the indentation of `defrecord` and `deftype` multiple airity protocol forms.
 
 ## 5.5.0 (2016-06-25)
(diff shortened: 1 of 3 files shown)


bbatsov:
This looks ok to me. I wonder why no one noticed the problem for so long. Ideally the change should be accompanied by some tests. You should also update the commit message to match our guidelines and add a changelog entry.

slipset:
@bbatsov if my commit-message is not up to par, please advice me on what's missing, as I'm a bit new to this.

Malabarba:
I think nobody noticed this because it apparently doesn't affect anything.
For instance, the backward/forward-sexp commands work fine (in fact, the test added by this PR already passes with the current code), and even running `(thing-at-point 'symbol)` successfully returns the whole symbols.

I agree it's good to fix the syntax table, but it would be nice to have a real test for this (one that fails now, but passes with the change).

bbatsov:
If `(thing-at-point 'symbol)` works properly by default than I'm wondering where the issue reported stems from...

PR #396, opened 2016-07-23
Fixing prefix syntaxes

vspinu:
I have altered both # and ? syntaxes and added a couple of test for that. Also moved syntax related tests into a dedicated file.


bbatsov:
Looks good to me. Thanks!

PR #397, opened 2016-07-23
[Fix #393] don't hard-code imenu-generic-expression

### LINKED ISSUE #393: Imenu matching is hardcoded
As far as I can tell [this](https://github.com/clojure-emacs/clojure-mode/blob/master/clojure-mode.el#L347-L349) usage is not the right way to go about it. 

As a result I cannot add my patterns to `imenu-generic-expression`.  

The line should probably be 

```elisp
(add-to-list 'imenu-generic-expression '(nil clojure-match-next-def 0))
```

vspinu:


bbatsov:
Looks good. Thanks!

PR #408, opened 2016-09-16
Use customizable prefix key

### LINKED ISSUE #407: Use customizable prefix key
## Expected behavior

The prefix key for `clojure-mode-map` should be customizable, as it is in packages like [Projectile](https://github.com/bbatsov/projectile/blob/master/projectile.el#L202).

## Actual behavior

The `C-c C-r` key sequence is [hard-coded](https://github.com/clojure-emacs/clojure-mode/blob/master/clojure-mode.el#L194).

Obvious.

## Environment & Version information

Not needed.

`clojure-mode`: 5.5.2

### Emacs version

Emacs: 24.5.50.1

### Operating system

OS: Fedora 14

xiongtx:
Fixes #407.

-----------------

- [ ] You've added tests (if possible) to cover your change(s). Indentation & font-lock tests are extremely important!

Thanks!

### DIFF
--- a/clojure-mode.el
+++ b/clojure-mode.el
@@ -191,38 +191,55 @@ Out-of-the box clojure-mode understands lein, boot and gradle."
           (and (listp value)
                (cl-every 'stringp value))))
 
+(defcustom clojure-mode-keymap-prefix "C-c C-r"
+  "Clojure mode keymap prefix."
+  :group 'clojure
+  :type 'string)
+
+(defvar clojure-mode-prefixed-bindings
+  '(("C-t" . clojure-thread)
+    ("t" . clojure-thread)
+    ("C-u" . clojure-unwind)
+    ("u" . clojure-unwind)
+    ("C-f" . clojure-thread-first-all)
+    ("f" . clojure-thread-first-all)
+    ("C-l" . clojure-thread-last-all)
+    ("l" . clojure-thread-last-all)
+    ("C-a" . clojure-unwind-all)
+    ("a" . clojure-unwind-all)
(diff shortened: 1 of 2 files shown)


clojure-mode.el
```diff
@@ -191,38 +191,55 @@ Out-of-the box clojure-mode understands lein, boot and gradle."
           (and (listp value)
                (cl-every 'stringp value))))
 
+(defcustom clojure-mode-keymap-prefix "C-c C-r"
```
bbatsov: The refactoring prefix shouldn't really be customizable in this manner. A major mode's keybindings should always start with `C-c`, no exceptions. Only the `C-r` part might be made customizable, but I don't see much value in doing this.

clojure-mode.el
```diff
@@ -191,38 +191,55 @@ Out-of-the box clojure-mode understands lein, boot and gradle."
...
+  "Clojure mode keymap prefix."
+  :group 'clojure
+  :type 'string)
+
+(defvar clojure-mode-prefixed-bindings
```
bbatsov: I'd define a nested keymap, instead of generating keybindings dynamically.

clojure-mode.el
Malabarba: Also, it should be called `clojure-mode-refactorings-keymap-prefix` or something like that. As only our refactoring commands are under this prefix.

xiongtx:
Closed due to [not being needed](https://github.com/clojure-emacs/clojure-mode/issues/407#issuecomment-247744686).

PR #415, opened 2016-11-01
Port let related refactorings from clj-refactor.el

benedekfazekas:
Migrate introduce let, move to let from clj-refactor.el. Add introduce
expanded let, forward slurp into let and backward slurp into let.

-----------------

Implementation follows the main outlines of the cljr code but is
reworked at certain places. Major differences are as follows:
- Expanded let is introduced: with a prefix argument let introduced N
lists up with all the occurrences of bound form replaced at addition
time.
- New function: slurp function into let form forward and backward. Added
value again is to replace bounded forms with their bound names in the
slurped forms. prefix argument can be used again to slurp multiple forms
into the let.
- Expand let is not ported from cljr. Instead `paredit-convolute-sexp`
is advised to replace forms with bound names when used on let like form.

Further notes:
- `string-trim` is moved upstream from cider (after merging this, cider
can be refactored to use the trim fns from `clojure-mode`)

Advise `paredit-convolute-sexp` when used on a let form as drop in replacement for `cljr-expand-let`.

### DIFF
--- a/clojure-mode.el
+++ b/clojure-mode.el
@@ -11,7 +11,7 @@
 ;; URL: http://github.com/clojure-emacs/clojure-mode
 ;; Keywords: languages clojure clojurescript lisp
 ;; Version: 5.6.0-cvs
-;; Package-Requires: ((emacs "24.3"))
+;; Package-Requires: ((emacs "24.4"))
 
 ;; This file is not part of GNU Emacs.
 
@@ -68,6 +68,7 @@
 (require 'imenu)
 (require 'newcomment)
 (require 'align)
+(require 'subr-x)
 
 (declare-function lisp-fill-paragraph  "lisp-mode" (&optional justify))
 
@@ -227,6 +228,10 @@ Out-of-the box clojure-mode understands lein, boot and gradle."
     (define-key map (kbd "n h") #'clojure-insert-ns-form-at-point)
     (define-key map (kbd "n u") #'clojure-update-ns)
     (define-key map (kbd "n s") #'clojure-sort-ns)
+    (define-key map (kbd "s i") #'clojure-introduce-let)
+    (define-key map (kbd "s m") #'clojure-move-to-let)
+    (define-key map (kbd "s f") #'clojure-let-forward-slurp-sexp)
+    (define-key map (kbd "s b") #'clojure-let-backward-slurp-sexp)
     map)
   "Keymap for Clojure refactoring commands.")
 (fset 'clojure-refactor-map clojure-refactor-map)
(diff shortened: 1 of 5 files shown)


benedekfazekas:
all yours ladies and gents for some good reviewing :)

clojure-mode.el
```diff
@@ -2061,6 +2108,209 @@ See: https://github.com/clojure-emacs/clj-refactor.el/wiki/cljr-cycle-if"
...
+  (unless n (setq n 1))
+  (dotimes (k n)
+    (save-excursion (clojure--let-forward-slurp-sexp-internal))))
+
+;;;###autoload
+(defun clojure-introduce-let (&optional n)
+  "Create a let form, binding the form at point.
+With a numeric prefix argument the let is introduced N lists up."
+  (interactive "P")
+  (clojure--introduce-let-internal (read-from-minibuffer "Name: ") n))
```
bbatsov: I guess we can have a better (clearer) prompt here.

clojure-mode.el
```diff
@@ -2061,6 +2108,209 @@ See: https://github.com/clojure-emacs/clj-refactor.el/wiki/cljr-cycle-if"
...
+
+;;;###autoload
+(defun clojure-move-to-let ()
+  "Move the form at point to a binding in the nearest let."
+  (interactive)
+  (clojure--move-to-let-internal (read-from-minibuffer "Name: ")))
```
bbatsov: Same here.

bbatsov:
You're welcome. :-)

clojure-mode.el
```diff
@@ -540,6 +558,23 @@ This could cause problems.
 It is simply `clojure-docstring-fill-prefix-width' number of spaces."
   (make-string clojure-docstring-fill-prefix-width ? ))
 
+;; moved upstream from cider
+(defun clojure-string-trim-left (string)
```
benedekfazekas: happy to do that. I will modify the PR accordingly

clojure-mode.el
```diff
@@ -440,20 +449,29 @@ ENDP and DELIMITER."
...
   (when (>= paredit-version 21)
     (let ((keymap (or keymap clojure-mode-map)))
       (define-key keymap "{" #'paredit-open-curly)
       (define-key keymap "}" #'paredit-close-curly))
     (add-to-list 'paredit-space-for-delimiter-predicates
                  #'clojure-space-for-delimiter-p)
     (add-to-list 'paredit-space-for-delimiter-predicates
-                 #'clojure-no-space-after-tag)))
+                 #'clojure-no-space-after-tag)
+    (advice-add 'paredit-convolute-sexp :after #'clojure--replace-let-bindings-and-indent)))
```
benedekfazekas: right, it is only available from 24.4

PR #424, opened 2017-01-27
Avoid throwing inside `clojure-match-next-def` on 0 arity def-like forms

### LINKED ISSUE #423: `clojure-match-next-def` doesn't support 0 arity top-level forms.
## Expected behavior

`clojure-match-next-def` shouldn't choke on this:

```clojure
(def-shortcuts)
```

## Actual behavior

`(forward-sexp)` is called [1]  when the cursor position is (denoted by `^`) `(def-shortcuts^)`, and this causes the following error:

```elisp
(scan-error "Containing expression ends prematurely" 9581 9582)
```

[1] https://github.com/clojure-emacs/clojure-mode/blob/0113aa969e09e31d65717d4a9c16c934c77dcb9b/clojure-mode.el#L667

Place your cursor at `^`:

```clojure
(def-shortcuts)^
```
invoke `(clojure-match-next-def)`. 

```
clojure-mode (version 5.7.0-snapshot)
```

```clojure-mode-20170120.2239```

### Emacs version

```
GNU Emacs 24.5.1 (x86_64-redhat-linux-gnu, GTK+ Version 3.18.9)
```

```
$ rpm -qi emacs
Name        : emacs
Epoch       : 1
Version     : 24.5
Release     : 10.fc23
Architecture: x86_64
... <snip> ...
Signature   : RSA/SHA256, Tue 12 Apr 2016 09:13:13 PM CDT, Key ID 32474cf834ec9c
ba
Source RPM  : emacs-24.5-10.fc23.src.rpm
Build Date  : Mon 11 Apr 2016 06:21:33 AM CDT
... <snip> ...
```

### Operating system

`Fedora 23 (Linux)`

### PR DESCRIPTION
**Pull request by emidln**
Fixes #423

### DIFF
--- a/clojure-mode.el
+++ b/clojure-mode.el
@@ -664,7 +664,9 @@ Called by `imenu--generic-function'."
         (down-list)
         (forward-sexp)
         (while (not found?)
-          (forward-sexp)
+          (condition-case nil
+              (forward-sexp)
+            (error nil))
           (or (if (char-equal ?[ (char-after (point)))
                               (backward-sexp))
                   (if (char-equal ?) (char-after (point)))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### Bugs fixed
+
+* Make `clojure-match-next-def` more robust against zero-arity def-like forms.
+
 ### New features
 
 * New interactive command `clojure-cycle-when`.

### CODE REVIEW

**Comment by bbatsov** [comment]
Looks good to me, but ideally this should have a unit test as well.

File: clojure-mode.el [hunk 1]
```diff
@@ -664,7 +664,9 @@ Called by `imenu--generic-function'."
         (down-list)
         (forward-sexp)
         (while (not found?)
-          (forward-sexp)
+          (condition-case nil
+              (forward-sexp)
```
Malabarba: `(ignore-errors (forward-sexp))` is shorter and does the same.

Malabarba:
Hi Brandon. Sorry we left this unreviewed for so long. =/
Indeed, it would be nice to have a unit test, specially since this is a bug fix. But if that's too much trouble I'm fine with accepting this without.

clojure-mode.el
bbatsov: True. :-)

PR #430, opened 2017-04-07
[Fix #429] Last occurrence sometimes not replaced for `move-to-let`

### LINKED ISSUE #429: Last occurrence of expression sometimes not replaced when using move-to-let
*Use the template below when reporting bugs. Please, make sure that
you're running the latest stable clojure-mode and that the problem you're reporting
hasn't been reported (and potentially fixed) already.*

**Remove all of the placeholder text in your final report!**

## Expected behavior

All occurrences of an expression inside the let should replaced with newly introduced bound name.

## Actual behavior

in case there are more than one occurrences of expression and move to let is not initiated from the last occurrence and actual bound name is longer than the expression being moved to let the last expression won't be replaced

benedekfazekas:
In case
* there are more than one occurrences of expression
* and move to let is not initiated from the last occurrence
* and actual bound name is longer than the expression being moved to let
the last expression won't be replaced.

Fix: end of `let` expression is not cached before calling `clojure--replace-sexps-with-binding`.

**Replace this placeholder text with a summary of the changes in your PR.
The more detailed you are, the better.**

-----------------

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Bugs fixed
 
+* [#429](https://github.com/clojure-emacs/clojure-mode/issues/429): Fix a bug causing last occurrence of expression sometimes is not replaced when using `move-to-let`.
 * [#423](https://github.com/clojure-emacs/clojure-mode/issues/423): Make `clojure-match-next-def` more robust against zero-arity def-like forms.
 
 ### New features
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Bugs fixed
 
+* [#429](https://github.com/clojure-emacs/clojure-mode/issues/429): Fix last occurrence of expression sometimes is not replaced when using `move-to-let`.
```
bbatsov: This sentence sounds a bit strange to me. :-) Perhaps "Fix a bug causing the last..."?

PR #431, opened 2017-05-02
Add defcustom for comment font-locking

arichiardi:
The patch adds a `clojure-comment-mode` defcustom that enable the user
to consider comments also the ones defined in a `(comment)` form.  The
default behavior is not changed, the comment reader macro is the only
one considered a comment.

The need started from @qqq on Slack, helped by @dpsutton. I always wanted the option, so I added it. I will update the docs if there are no objections (to verify the behavior use `font-lock-fontify-block` in emacs after setting the `clojure-comment-mode` with `ielm`).

-----------------

- [X] The commits are consistent with our [contribution guidelines][1]
- [ ] You've added tests (if possible) to cover your change(s). Indentation & font-lock tests are extremely important!
- [ ] All tests are passing (`make test`)
- [X] The new code is not generating bytecode or `M-x checkdoc` warnings
- [ ] You've updated the changelog (if adding/changing user-visible functionality)

### DIFF
--- a/clojure-mode.el
+++ b/clojure-mode.el
@@ -614,15 +614,27 @@ If JUSTIFY is non-nil, justify as well as fill the paragraph."
   "Regexp matching the start of a comment sexp.
 The beginning of match-group 1 should be before the sexp to be
 marked as a comment.  The end of sexp is found with
-`clojure-forward-logical-sexp'.
+`clojure-forward-logical-sexp'.")
 
-By default, this only applies to code after the `#_' reader
-macro.  In order to also font-lock the `(comment ...)' macro as a
-comment, you can set the value to:
-    \"#_ *\\\\(?1:[^ ]\\\\)\\\\|\\\\(?1:(comment\\\\_>\\\\)\"")
+(defvar clojure--reader-and-comment-regexp
+  "#_ *\\(?1:[^ ]\\)\\|\\(?1:(comment\\_>\\)"
+  "Regexp matching both `#_' macro and a comment sexp." )
+
+(defcustom clojure-comment-regexp clojure--comment-macro-regexp
+  "Comment mode.
+
+The possible values for this variable are keywords indicating
+what is considered a comment (affecting font locking).
+
+    - Reader macro `#_' only - the default
+    - Reader macro `#_' and `(comment)'"
+  :type '(choice (const :tag "Reader macro `#_' and `(comment)'" clojure--reader-and-comment-regexp)
+                 (other :tag "Reader macro `#_' only" clojure--comment-macro-regexp))
+  :package-version '(clojure-mode . "5.7.0"))
 
 (defun clojure--search-comment-macro-internal (limit)
-  (when (search-forward-regexp clojure--comment-macro-regexp limit t)
+  "Search for a comment forward stopping at LIMIT."
+  (when (search-forward-regexp clojure-comment-regexp limit t)
     (let* ((md (match-data))
            (start (match-beginning 1))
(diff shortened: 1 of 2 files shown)


bbatsov:
Those are not really comments and I think adding such functionality is misleading. That's a form that will return `nil` in the end... We used to have this and I deleted it a while ago, as I felt it conveyed the wrong idea.

bbatsov:
Anyways, I'll think a bit about this, but the fact the existing code made it easy enough to support this for those who really wanted it is good enough for me.

arichiardi:
@bbatsov I agree with you totally, it can be misleading for beginners, that is why it is off by default :smile: It is totally worth having though for folks.

Malabarba:
I think it's a bit of overengineering to add a defcustom just to decide between two defvars. Instead, we could convert the defvar we already have into a defcustom, and configure the two possible values in its `:type`. Then we wouldn't need to change the function (although your change is pretty small and reasonable).

Malabarba:
By the way, there's a third value we could use too. If we use the following regexp, I think it will font lock only the *contents* of the `comment` macro, but not the macro itself. That might be more in line with the idea that the macro itself is actually meaningful to the code, but its contents are not. 

 `"#_ *\\(?1:[^ ]\\)\\|(comment\\_>\\(?1:\\)"`

arichiardi:
Oh ok I did that because it was the only way I knew 😀 do you have an example with `:type`? Sounds better for sure.
About the content, I agree it is probably better!

bbatsov:
@arichiardi Please, make the changes suggested by @Malabarba.

arichiardi:
Oh I forgot about this. For sure 😀

arichiardi:
I have changed things here, I don't know if the `choice` still works there, I have strange behaviors in `customize-mode` here.

Also, the last regexp does not work when there are multiple forms inside `comment`:
2017-06-10-191824_204x55_scrot

Not a regex master here, will try my best. Ping @Malabarba :smile:

clojure-mode.el
```diff
@@ -614,15 +614,28 @@ If JUSTIFY is non-nil, justify as well as fill the paragraph."
...
+
+(defcustom clojure-comment-regexp clojure--comment-macro-regexp
+  "Comment mode.
+
+The possible values for this variable are keywords indicating
+what is considered a comment (affecting font locking).
+
+    - Reader macro `#_' only - the default
+    - Reader macro `#_' and `(comment)'
+    - Reader macro `#_' content and `(comment)'"
```
Malabarba: 3 items are listed by only 2 are implemented below.

Malabarba:
> Also, the last regexp does not work when there are multiple forms inside comment:

True. Unfortunately, that can't be fixed with regexps.
That might be fixed if you change `(match-beginning 1)` to `(match-beginning 0)` and add a `(forward-char -1)` after [this line](https://github.com/clojure-emacs/clojure-mode/pull/431/commits/bbe555a5e5c876b202e57e7ec4b0c430dd7d99ef#diff-50632fdcf67b6d3c8505a6aa071a2bd0R647).

However, I just noticed a bug with `clojure-forward-logical-sexp` that breaks that. `clojure-forward-logical-sexp` should skip over the `#_` like it does with other reader macros (like `#break`), but for some reason it's treating the `#_` like an entire sexp.

clojure-mode.el
arichiardi: Will fix thanks!

arichiardi:
Ok thanks that you tried! I will leave it out for now then. The feature itself is quite nice.

PR #432, opened 2017-05-21
Include defmethod dispatch key in imenu items

vspinu:
All`(defmethod foo dispatch-key ...)` currently show as `foo` in imenu. Fix that by including corresponding `dispatch-key` into imenu items.

I have also replaced cl-destructuring-bind with let as it doesn't work with edebug in emacs 25.2.2. Cannot dig into that at the moment.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ### New features
 
+* Add imenu support for multimethods
 * New interactive command `clojure-cycle-when`.
 * New interactive command `clojure-cycle-not`.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ### New features
 
 * Add imenu support for multimethods
+* Make imenu recognize indented def-forms
 * New interactive command `clojure-cycle-when`.
 * New interactive command `clojure-cycle-not`.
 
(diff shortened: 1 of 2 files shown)


clojure-mode.el
```diff
@@ -657,23 +657,30 @@ and `(match-end 1)'."
...
             (forward-sexp))
-          (or (if (char-equal ?[ (char-after (point)))
-                              (backward-sexp))
-                  (if (char-equal ?) (char-after (point)))
+          (or (when (char-equal ?[ (char-after (point)))
+                (backward-sexp))
+              (when (char-equal ?) (char-after (point)))
                 (backward-sexp)))
-          (cl-destructuring-bind (def-beg . def-end) (bounds-of-thing-at-point 'sexp)
+          (let* ((val (bounds-of-thing-at-point 'sexp))
```
bbatsov: This change should ideally be a separate commit.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 
 ### New features
 
+* Imenu understands defmethods
```
bbatsov: Added imenu support for multimethods.

vspinu:
Fixed.

vspinu:
I am adding one more commit to make imenu recognize indented def-forms. It's not uncommon to define  within a let block or a comment.

PR #444, opened 2017-08-07
[Fix #443] Don't consider #? as starting non-logical sexp

### LINKED ISSUE #443: clojure-backward-logical-sexp jumps over conditional `#?` sexps
## Actual behavior

```clj
(def tt 1)

#?(:cljs
   (defn foo []))

#?(:cljs
   (defn bar []))

(defn bar [])
| ;; <- point here
```

`M-x clojure-backward-logical-sexp` will land just before first `#?(:cljs`.  This breaks eval of defuns in cider. 

The issue is with how `clojure-backward-logical-sexp` determines non-logical sexps. That simple regexp logic breaks there.

vspinu:
Fixing this in a simplest way I can think of. Judging from the docs of `clojure--looking-at-non-logical-sexp` `#?` should not even be considered as non-logical sexps under any circumstances.

I think `?` in that regexp is some kind of legacy remnant. `#?@` are now marked as prefix chars  and `backward-sexp` does the right thing on prefixed sexps.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Bugs fixed
 
 * Fix fill-paragraph in multi-line comments.
+* [#443](https://github.com/clojure-emacs/clojure-mode/issues/443): Fix behavior of `clojure-forward-logical-sexp` and `clojure-backward-logical-sexp` with conditional macros.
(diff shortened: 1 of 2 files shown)


vspinu:
Travis test broke for unrelated reason.

Malabarba:
Merged! Thanks a ton.

bbatsov:
I was just going to say this needs to have a regression unit test. :-)

Issue #241, opened 2014-07-18
Identation issues in docstrings


expez:
```clojure
(ns test.core
  "If I write a really long line without worrying about indentation
  and then call clojure-fill-paragraph to indent once I'm done it
  looks like this and all is well.

This paragraph on the other hand was created by hitting RET to indent
  manually.  Every time I do this the indentation moves all the way to
  the left, instead of staying aligned.  Calling clojure-fill-paragraph
  now will indent all but the first line in this paragraph properly.")
```

tsdh:
@bzg The problem I have with the current `clojure-indent-line` function is that I frequently have example code and ASCII art in docstrings which shall not be touched.  E.g., I have

```` clojure
(defn foo
  "foo should be called like

     (with-bar something
       (foo x))

  where something is something."
  [x]
  (do-stuff-with m))
````

and after indentation I end up with

```` clojure
(defn foo
  "foo should be called like

  (with-bar something
  (foo x))

  where something is something."
  [x]
  (do-stuff-with m))
````

Of course, now you'll say "then don't hit TAB in docstrings" but I use [aggressive-indent-mode](https://github.com/Bruce-Connor/aggressive-indent-mode) which re-indents while typing.

Right now, I simply `fset` `clojure-indent-line` to `lisp-indent-line` which returns `'noindent` for strings.  That solves my issue but I don't think that forcing every line in a docstring to start with the same number of spaces is a good feature in general.  Maybe adding the clojure fill prefix in case there are less spaces would be ok, but removing spaces a user has intentionally added is not IMHO.

bzg:
Hi Tassilo,

I don't have enough time at hand right now -- if you can come up with a patch for this, I'll gladly accept it!

Thanks in advance.

stig:
I'm having indentation issues on things like this: https://github.com/Datomic/day-of-datomic/blob/master/src/datomic/samples/schema.clj#L50

sandhu:
The #289 PR that got merged this weekend should address this issue.

tsdh:
Great, thanks!

stig:
Hurrah!

raxod502:
But it still looks like `electric-indent-mode` is broken? I.e. if I press `RET` in a Clojure docstring, then point is at column 0 until I press `TAB` at which point it moves to column 2.

bzg:
The behavior I see in a docstring with `clojure-mode version 5.7.0-snapshot` is this one:

[...]

PR #447, opened 2017-10-01
clojure-mode.el (clojure-mode): Force elec. indentation in a docstring

bzg:
By default, `electric-layout-post-self-insert-function` prevents indentation in strings and comments.  When `electric-indent-mode` is on, we want `newline` to reindent the line when the point is within a docstring, so this patch adds a hook to `electric-indent-functions` to that effect.

See https://github.com/clojure-emacs/clojure-mode/issues/241#issuecomment-332323785 for a discussion about this bug.

### DIFF
--- a/clojure-mode.el
+++ b/clojure-mode.el
@@ -525,8 +525,10 @@ replacement for `cljr-expand-let`."
   (clojure-mode-variables)
   (clojure-font-lock-setup)
   (add-hook 'paredit-mode-hook #'clojure-paredit-setup)
+  ;; `electric-layout-post-self-insert-function' prevents indentation in strings
+  ;; and comments, force indentation in docstrings:
   (add-hook 'electric-indent-functions
-            (lambda (c) (if (clojure-in-docstring-p) 'do-indent))))
+            (lambda (char) (if (clojure-in-docstring-p) 'do-indent))))
 
 (defcustom clojure-verify-major-mode t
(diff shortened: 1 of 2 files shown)


raxod502:
It appears to work for me. Thanks!

clojure-mode.el
```diff
@@ -524,7 +524,9 @@ replacement for `cljr-expand-let`."
 \\{clojure-mode-map}"
   (clojure-mode-variables)
   (clojure-font-lock-setup)
-  (add-hook 'paredit-mode-hook #'clojure-paredit-setup))
+  (add-hook 'paredit-mode-hook #'clojure-paredit-setup)
+  (add-hook 'electric-indent-functions
```
bbatsov: I'd really add some explanations here, as people are going to be confused by this bit of code for sure. 

You might also consider adding some simple unit test verifying the correct behavior when electric indent is enabled.

bbatsov:
Apart from my small inline remark - this needs a changelog entry.

bzg:
Hi @bbatsov, thanks for the feedback! I've completely the previous commit with an inline comment and a changelog entry. Hopefully I've done it right. Thanks!

abo-abo:
This commit breaks variable docstrings indentation for me.

    |(def foo "docstring" 123)

Entering a char in the docstring forces an indent by two spaces:

    |  (def foo "docstring1" 123)

I had to disable `electric-indent-mode` in my config.

PR #452, opened 2017-10-06
[Fix #451] respect project root dir suggested by custom function

### LINKED ISSUE #451: project root dir is ignored when used with projectile
This can be painful when clojure-mode is used together with projectile and cider. For certain operations (like jack-in to the REPL) cider tries to find a project root dir via `clojure-project-dir` and in most scenarios it works. The problem however appears when project uses few other build tools at the same time.

Imagine following directory structure:
```
/proj
|-> build.boot
|-> subproject_1
      |-> build.gradle
```
Projectile allows to customize a way how the project root dir is calculated and having `build.gradle` files ignored in this case it locates `/proj` root with `build.boot` inside correctly. Unfortunately, as `clojure-project-dir` doesn't use that "knowledge" it simply forces subproject_1 as a root dir (when called within this directory) and jack-in goes with gradle.

I prepared a simple PR which mitigates this problem. Please, take a look at it and accept if it makes sense.

mbuczko:
This is a humble attempt to get clojure-mode respecting project root directory suggested by a custom function (like projectile-project-root), as described in #451

### DIFF
--- a/clojure-mode.el
+++ b/clojure-mode.el
@@ -192,7 +192,7 @@ Out-of-the box `clojure-mode' understands lein, boot and gradle."
           (and (listp value)
                (cl-every 'stringp value))))
 
-(defcustom clojure-project-root-function 'clojure-project-dir
+(defcustom clojure-project-root-function 'clojure-project-root-path
   "Function to locate clojure project root directory."
   :type 'function
   :risky t
@@ -1604,6 +1604,13 @@ nil."
 
 
 (defun clojure-project-dir (&optional dir-name)
+  "Return the absolute path to the project's root directory.
+
+Call is delegated down to `clojure-project-root-function' with
+optional DIR-NAME as argument."
+  (funcall clojure-project-root-function dir-name))
+
+(defun clojure-project-root-path (&optional dir-name)
   "Return the absolute path to the project's root directory.
 
 Use `default-directory' if DIR-NAME is nil.
--- a/clojure-mode.el
+++ b/clojure-mode.el
@@ -1620,10 +1620,8 @@ Return nil if not inside a project."
                         (mapcar (lambda (fname)
                                   (locate-dominating-file dir-name fname))
                                 clojure-build-tool-files))))
-    (or (ignore-errors
-          (funcall clojure-project-root-function))
-        (when (> (length choices) 0)
-          (car (sort choices #'file-in-directory-p))))))
+    (when (> (length choices) 0)
+      (car (sort choices #'file-in-directory-p)))))
 
 (defun clojure-project-relative-path (path)
   "Denormalize PATH by making it relative to the project root."
--- a/clojure-mode.el
+++ b/clojure-mode.el
@@ -192,6 +192,11 @@ Out-of-the box `clojure-mode' understands lein, boot and gradle."
           (and (listp value)
                (cl-every 'stringp value))))
 
+(defcustom clojure-project-root-locating-function nil
+  "Alternative function to locate clojure project root directory, eg. projectile-project-root"
+  :type 'symbol
+  :safe 'symbolp)
+
 (defcustom clojure-refactor-map-prefix (kbd "C-c C-r")
   "Clojure refactor keymap prefix."
   :type 'string
@@ -1607,9 +1612,9 @@ Return nil if not inside a project."
                         (mapcar (lambda (fname)
                                   (locate-dominating-file dir-name fname))
                                 clojure-build-tool-files))))
-    (or (and (fboundp 'projectile-project-root)
+    (or (and (fboundp clojure-project-root-locating-function)
              (condition-case err
-                 (projectile-project-root)
+                 (funcall clojure-project-root-locating-function)
                (error nil)))
         (when (> (length choices) 0)
           (car (sort choices #'file-in-directory-p))))))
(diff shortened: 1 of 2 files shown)


clojure-mode.el
```diff
@@ -1607,8 +1612,12 @@ Return nil if not inside a project."
                         (mapcar (lambda (fname)
                                   (locate-dominating-file dir-name fname))
                                 clojure-build-tool-files))))
-    (when (> (length choices) 0)
-      (car (sort choices #'file-in-directory-p)))))
+    (or (and (fboundp clojure-project-root-locating-function)
+             (condition-case err
```
Malabarba: This is just ignore-errors

clojure-mode.el
```diff
@@ -192,6 +192,11 @@ Out-of-the box `clojure-mode' understands lein, boot and gradle."
           (and (listp value)
                (cl-every 'stringp value))))
 
+(defcustom clojure-project-root-locating-function nil
+  "Alternative function to locate clojure project root directory, eg. projectile-project-root"
+  :type 'symbol
```
Malabarba: I think there's a type: 'function

clojure-mode.el
```diff
@@ -192,6 +192,11 @@ Out-of-the box `clojure-mode' understands lein, boot and gradle."
...
+  :safe 'symbolp)
```
Malabarba: This variable is funcalled, so no value is safe for it. In fact, we should probably mark it as risky.

clojure-mode.el
```diff
@@ -192,6 +192,11 @@ Out-of-the box `clojure-mode' understands lein, boot and gradle."
           (and (listp value)
                (cl-every 'stringp value))))
 
+(defcustom clojure-project-root-locating-function nil
```
bbatsov: I'd name this just `clojure-project-root-function`.

clojure-mode.el
```diff
@@ -192,6 +192,11 @@ Out-of-the box `clojure-mode' understands lein, boot and gradle."
...
+  "Alternative function to locate clojure project root directory, eg. projectile-project-root"
```
bbatsov: Why alternative - let's just set it by default to the currently existing function.

clojure-mode.el
```diff
@@ -1607,8 +1612,12 @@ Return nil if not inside a project."
                         (mapcar (lambda (fname)
                                   (locate-dominating-file dir-name fname))
                                 clojure-build-tool-files))))
-    (when (> (length choices) 0)
-      (car (sort choices #'file-in-directory-p)))))
+    (or (and (fboundp clojure-project-root-locating-function)
```
bbatsov: I'd simply remove those check here - if the user set something explicitly we can assume they know what they are doing. Fallbacks are pointless and confusing here.

clojure-mode.el
bbatsov: The defcustom should also have a `:package-version` property pointing this was added in version 5.7.0.

vspinu:
Shouldn't this be using `project-find-functions` (new emacs 25 functionality in `project.el`)?  If users can already add hooks to `project-find-functions`, then this functionality seems redundant.

bbatsov:
> Shouldn't this be using project-find-functions (new emacs 25 functionality in project.el)? If users can already add hooks to project-find-functions, then this functionality seems redundant.

In theory - yes. In practice this is more or less a prototype that wasn't really polished. I was actually discussing this with its author just a couple of days ago. He wanted us to collaborate on providing more extension points for Projectile, but frankly I don't really see much point in working towards this universal API given the fact I'm not really interested in using it and apart from a few generic functions each project management library has way too many specifics to it. On the other side - this could allow a lot of modes to drop at least the product root detection code they cart along.

mbuczko:
@bbatsov I applied your suggestions but to get rid of unnecessary fallbacks and set `clojure-project-root-function` customization by default to the function currently used to calculate project root path I had to split `clojure-project-dir` into 2 parts. I'm not sure if that's done right, any comments nicely welcomed :)

bbatsov:
Why did you split it in two? I don't see any reason to do, even after taking a look at the code.

mbuczko:
This was to set `clojure-project-root-function` to default value (function) which would call what we have right now in clojure-mode to calculate root-path. I couldn't set it up simply to `clojure-project-dir` as having my funcall inside, this function would be called in recursive loop (funcall would call `clojure-project-dir` again which in turn would call funcall and so on...).

bbatsov:
Ah, yeah - the name you have to the old function confused me a bit. This looks ok overall.

clojure-mode.el
```diff
@@ -192,6 +192,12 @@ Out-of-the box `clojure-mode' understands lein, boot and gradle."
           (and (listp value)
                (cl-every 'stringp value))))
 
+(defcustom clojure-project-root-function 'clojure-project-root-path
```
Malabarba: Change the ' to a #'

Malabarba:
Besides that last comment, it looks good to me.

bbatsov:
It still needs a changelog entry.

mbuczko:
Changelog updated.

dgutov:
> this could allow a lot of modes to drop at least the product root detection code they cart along

It would allow e.g. `M-x project-find-regexp` to use Projectile. And some other functions in core Emacs.

PR #455, opened 2017-10-17
Fix namespace font locking

Bost:

I added only some visual tests in the test.clj. I'll add some proper ert tests in the next iteration on this PR.

- [ ] All tests are passing (`make test`)

Using cask for testing is a pain :-(
1. Cask installation doesn't work from behind a proxy
2. Being behind or a proxy or not - in both cases I get:
```
$ make test
make: *** No rule to make target '/home/bost/dev/clojure-mode/.cask/25.3/elpa', needed by 'test'.  Stop.
```
`M-x spacemacs/ert-run-tests-buffer` shows no new errors (on top of those 4 existing already):
```
Selector: (satisfies (lambda (test) (eq cbuf (spacemacs//find-ert-test-buffer test))))
Passed:  30
Failed:  4 (4 unexpected)
Skipped: 0
Total:   34/34

Started at:   2017-10-17 17:44:11+0200
Finished.
Finished at:  2017-10-17 17:44:13+0200

F.F.......F.............F.........

F clojure-mode-syntax-table/characters
    (ert-test-failed
     ((should
       (eq
        (clojure-test-face-at 1 2 "\\1")
        'clojure-character-face))
      :form
      (eq various-faces clojure-character-face)
      :value nil))

F clojure-mode-syntax-table/comment-macros
    (ert-test-failed
     ((should
       (equal
        (clojure-test-face-at 5 41 "#_ 
;; some crap
 (lala 0101
 lao

 0 0i)")
        'font-lock-comment-face))
      :form
      (equal various-faces font-lock-comment-face)
      :value nil :explanation
      (different-atoms various-faces font-lock-comment-face)))

F clojure-mode-syntax-table/fn
    (ert-test-failed
     ((should
       (eq
        (clojure-test-face-at 2 3)
        'font-lock-keyword-face))
      :form
      (eq t font-lock-keyword-face)
      :value nil))

F clojure-mode-syntax-table/ns-macro
    (ert-test-failed
     ((should
       (eq
        (clojure-test-face-at 1 10 "[ns name]")
        nil))
      :form
      (eq various-faces nil)
      :value nil))
```

I fixed a bug not previously reported. No readme change is needed.


Malabarba:
Thanks. I'll try to figure out this week why our tests are failing on master.
Meanwhile, could you be a little more specific in the ChangeLog message? Just giving an example of a namespace that was fixed would be great.

clojure-mode.el
```diff
@@ -855,19 +855,19 @@ any number of matches of `clojure--sym-forbidden-rest-chars'."))
       ;; Java interop highlighting
       ;; CONST SOME_CONST (optionally prefixed by /)
       ("\\(?:\\<\\|/\\)\\([A-Z]+\\|\\([A-Z]+_[A-Z1-9_]+\\)\\)\\>" 1 font-lock-constant-face)
+      ;; special handling for foo.bar.baz in the namespace definition
```
bbatsov: I'm a bit confused by this comment. What exactly is special here?

test.clj
```diff
@@ -1,7 +1,22 @@
...
+   [clojure.something]
+   [something.s]
+   [oneword]
+   ;; TODO mixedCase shouldn't be font-locked to clojure-interop-method-face
+   [mixedCase]
+   [CamelCase]))
+
+;; examples of valid namespace definitions
+(comment
+  (ns .validns)
```
bbatsov: All of those should ideally be unit tests.

Bost:
Well, if the clojure allows it then we should support it.

Le 31 oct. 2017 10:35, "Bozhidar Batsov" <notifications@github.com> a
écrit :

bbatsov:
I'm mostly concerned how this is going to impact the font-locking of other things. As no one has asked to support such esoteric namespaces I'm wondering what exactly do we gain from doing so - are solving a real problem or just creating one?

Bost:
> I'm mostly concerned how this is going to impact the font-locking of other things. As no one has asked to support such esoteric namespaces I'm wondering what exactly do we gain from doing so - are solving a real problem or just creating one?

Not just esoteric, even simple namespaces:
```
(ns Foo-bar)
(ns Foo.bar)
(ns foo.Bar.baz)
```
don't get proper font-locking. I repeat, if clojure allows it then we should feel obliged to support it. And if we're not happy with some of its properties (which in itself is perfectly fine), then we should discuss our concerns with the authors. We may write down some style guide rules or recommendations but not more than that. Besides, where exactly is the border between "esoteric" and "usual"? And also what about a machine generated code - macro expansions, clojure.spec test cases etc.? Shouldn't it be correctly font-locked, too?

pdbrown:
Slightly off-topic, but running `make elpa` before `make test` should fix the `No rule to make target ... needed by 'test'`. The Makefile could be changed `s/test: $(PKGDIR)/test: elpa/`, but I'm not sure if that's always ok.

clojure-mode.el
Bost: Good question. It might be better to either: (A) write here:
```
      ;; special handling for namespace definition
```
or (B) remove the comment in the first place.

And BTW I have a feeling the lines 865 to 870:
```
        ;; Foo Bar$Baz Qux_ World_OpenUDP Foo. Babylon15.
        ("\\(?:\\<\\|\\.\\|/\\|#?^\\)\\([A-Z][a-zA-Z0-9_]*[a-zA-Z0-9$_]+\\.?\\>\\)" 1 font-lock-type-face)
        ;; foo.bar.baz
        ("\\<^?\\([a-z][a-z0-9_-]+\\.\\([a-z][a-z0-9_-]*\\.?\\)+\\)" 1 font-lock-type-face)
```
are kind of a redundant here but it would be a too much of a change in one PR (at least for me) so I rather leave it untouched.

test.clj
Bost: I know, I know. Patience please :)

clojure-mode.el
bbatsov: Yeah, that's probably true. I think there are some tests for them, so if you remove them you'd know pretty soon. :-)

test.clj
Bost: Done. See the e5cc4d0 and a581bba.

bbatsov: No problem. Just let me know when you get to this.

bbatsov:
> don't get proper font-locking. I repeat, if clojure allows it then we should feel obliged to support it. And if we're not happy with some of its properties (which in itself is perfectly fine), then we should discuss our concerns with the authors. We may write down some style guide rules or recommendations but not more than that. Besides, where exactly is the border between "esoteric" and "usual"? And also what about a machine generated code - macro expansions, clojure.spec test cases etc.? Shouldn't it be correctly font-locked, too?

Fair enough.

test/clojure-mode-font-lock-test.el
```diff
@@ -169,6 +169,9 @@ POS."
 
 (ert-deftest clojure-mode-syntax-table/namespaced-symbol ()
   :tags '(fontification syntax-table)
+  (should (eq (clojure-test-face-at 1 5 "one.p") 'font-lock-type-face))
```
bbatsov: This seems unrelated to the rest of the test.

Bost: I'm closing this PR and going to send you a new PR where I fixed this point.

PR #458, opened 2017-10-31
Move to top level before `re-search-backward` in `clojure-find-ns`

xiongtx:
Moving to top level avoids improper matching behavior due to being in middle
of match.

Fix clojure-emacs/cider#2100

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Bugs fixed
 
+* [#458](https://github.com/clojure-emacs/clojure-mode/pull/458): Get correct ns when in middle of ns form with `clojure-find-ns`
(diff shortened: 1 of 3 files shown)


bbatsov:
The solution seems reasonable to me. Just update the changelog and add a couple of unit tests covering the fix. The CI has been broken for a while for some mysterious reason, but the tests are still runnable locally.

xiongtx:
> CI has been broken for a while for some mysterious reason...

Strange indeed. Posted a [question](https://emacs.stackexchange.com/questions/36510/package-lacks-a-version-or-package-version-header-in-file) on the Emacs StackExchange.

bbatsov:
> Strange indeed. Posted a question on the Emacs StackExchange.

My guess is this is something to do with Cask, as the error message is from `package.el` itself, but I've been super busy lately and just ignored this for now. 

Thanks for the great PR!

PR #460, opened 2017-11-01
Add font-lock for new vars in 1.9. Fixes #459

### LINKED ISSUE #459: Add font-lock for new vars in 1.9
clojure-1.9 introduces a bunch of new vars, eg pos-int?.
It would be nice to have font-locking support for these.

slipset:


bbatsov:
This should also be mentioned in the changelog.

PR #462, opened 2017-11-06
No special font-locking for mixedCase, CamelCase

Bost:
This is an alternative approach to my PR regarding fix-namespace-font-locking.
I find out the clojure-interop-method-face is not really needed for mixedCase, CamelCase... and it simplifies the regexp a lot. What do you think about that?
In case you think the mixedCase, CamelCase are needed indeed, then I'd build-up/rework their pattern matching based on this simplification.


bbatsov:
Perhaps this is the better approach indeed. We'll lose the special font-locking for interop methods, but it didn't seem particularly important to begin with.

test/clojure-mode-font-lock-test.el
```diff
@@ -161,7 +161,7 @@ POS."
     (should (eq (clojure-test-face-at 2 3) 'font-lock-type-face))
     (should (eq (clojure-test-face-at 5 14) 'font-lock-type-face))))
 
-(ert-deftest clojure-mode-syntax-table/namespace ()
+(ert-deftest clojure-mode-syntax-table/segmented-symbol ()
```
bbatsov: It seems to me that those tests should actually fail now - from what I gather by your changes we aim to font-lock as ns-es only ns symbols that are in the `ns` macro.

Bost: > Perhaps this is the better approach indeed. 

Glad to hear it :)

> It seems to me that those tests should actually fail now - from what I gather by your changes we aim to font-lock as ns-es only ns symbols that are in the ns macro.

Yea, I'll work on it. (I just wanted to hear your general opinion in the first place.) I'll tell you when have something to show.

Bost:
I made a good progress. Please have a look. Thx.

Bost:
I made a good progress. Please have a look. Thx.

clojure-mode.el
```diff
@@ -847,29 +842,40 @@ any number of matches of `clojure--sym-forbidden-rest-chars'."))
...
-      ("\\(?:\\<\\|/\\)\\([A-Z]+\\|\\([A-Z]+_[A-Z1-9_]+\\)\\)\\>" 1 font-lock-constant-face)
-      ;; .foo .barBaz .qux01 .-flibble .-flibbleWobble
-      ("\\<\\.-?[a-z][a-zA-Z0-9]*\\>" 0 'clojure-interop-method-face)
-      ;; Foo Bar$Baz Qux_ World_OpenUDP Foo. Babylon15.
-      ("\\(?:\\<\\|\\.\\|/\\|#?^\\)\\([A-Z][a-zA-Z0-9_]*[a-zA-Z0-9$_]+\\.?\\>\\)" 1 font-lock-type-face)
-      ;; foo.bar.baz
-      ("\\<^?\\([a-z][a-z0-9_-]+\\.\\([a-z][a-z0-9_-]*\\.?\\)+\\)" 1 font-lock-type-face)
-      ;; (ns namespace) - special handling for single segment namespaces
+
       (,(concat "(\\<ns\\>[ \r\n\t]*"
```
bbatsov: Add some comment for each regular expression, otherwise it's hard to understand their purpose.

clojure-mode.el
```diff
@@ -847,29 +842,40 @@ any number of matches of `clojure--sym-forbidden-rest-chars'."))
...
                 "\\(?:\\^?{[^}]+}[ \r\n\t]*\\)*"
                 ;; namespace
-                "\\([a-z0-9-]+\\)")
-       (1 font-lock-type-face nil t))
-      ;; fooBar
-      ("\\(?:\\<\\|/\\)\\([a-z]+[A-Z]+[a-zA-Z0-9$]*\\>\\)" 1 'clojure-interop-method-face)
+                "\\(" clojure--sym-regexp "\\)")
+       (1 font-lock-type-face))
+
+      (,(concat "\\(:\\)\\(" clojure--sym-regexp "\\)\\(/\\)\\(" clojure--sym-regexp "\\)")
```
bbatsov: I'm quite puzzled by this regexp.

clojure-mode.el
```diff
@@ -847,29 +842,40 @@ any number of matches of `clojure--sym-forbidden-rest-chars'."))
...
+       (1 'clojure-keyword-face)
+       (2 font-lock-type-face)
+       (3 'default)
+       (4 'clojure-keyword-face))
+      (,(concat "\\(:\\)\\(" clojure--sym-regexp "\\)")
+       (1 'clojure-keyword-face)
+       (2 'clojure-keyword-face))
+
+      ;; type-hints
+      (,(concat "\\(#^\\)\\(" clojure--sym-regexp "\\)\\(/\\)\\(" clojure--sym-regexp "\\)")
```
bbatsov: I guess the first is some qualified symbol, but certainly some examples would make the intent clearer.

bbatsov:
I see 3 tests are failing on Travis. This also needs a changelog entry.

test.clj
```diff
@@ -1,7 +1,38 @@
...
-            [something.s]))
+  (:require
+   [oneword]
+   [seg.mnt]
+   [mxdCase]
+   [CmlCase]
+   [veryCom|pLex.stu-ff]))
+
+;; TODO the face of 'fn' is 't'. Where is it defined?
+(fn foo [x] x)
```
bbatsov: `fn` is in the list of special forms, so it should be font-locked as one. I can't imagine what can cause the problem you've mentioned. At least I don't see anything obvious in your changes.

clojure-mode.el
```diff
@@ -847,29 +842,40 @@ any number of matches of `clojure--sym-forbidden-rest-chars'."))
...
       ;; Character literals - \1, \a, \newline, \u0000
       ("\\\\\\([[:punct:]]\\|[a-z0-9]+\\>\\)" 0 'clojure-character-face)
-      ;; foo/ Foo/ @Foo/ /FooBar
-      (,(concat "\\(?:\\<:?\\|\\.\\)@?\\(" clojure--sym-regexp "\\)\\(/\\)")
-       (1 font-lock-type-face) (2 'default))
-      ;; Constant values (keywords), including as metadata e.g. ^:static
-      ("\\<^?\\(:\\(\\sw\\|\\s_\\)+\\(\\>\\|\\_>\\)\\)" 1 'clojure-keyword-face append)
-      ;; Java interop highlighting
-      ;; CONST SOME_CONST (optionally prefixed by /)
-      ("\\(?:\\<\\|/\\)\\([A-Z]+\\|\\([A-Z]+_[A-Z1-9_]+\\)\\)\\>" 1 font-lock-constant-face)
```
bbatsov: And I guess no special font-locking for SCREAMING_UPPER_CASE.

clojure-mode.el
Bost: See the test.clj

clojure-mode.el
Bost: Comment added.

clojure-mode.el
Bost: Comment added.

test.clj
Bost: I need to have a look at it. Give me some time please.

clojure-mode.el
Bost: Right, no special font-locking for SCREAMING_UPPER_CASE at the moment. But feel free to change or propose something else.

bbatsov: I'm just mentioning it. I don't feel strongly about this.

clojure-mode.el
bbatsov: I meant there should be some examples here as well, so it's easier to map those regular expressions to what they are supposed to font-lock.

bbatsov:
@Bost How is all of this shaping up?

clojure-mode.el
Bost: > I meant there should be some examples here

(simple) examples added. More elaborate examples are in the test.clj

Bost:
> I see 3 tests are failing on Travis. This also needs a changelog entry.

I'm not sure how to document it. I just (partially) fixed one of them, so:
- `clojure-find-ns-test` fails only in emacs-24.4-travis, emacs-24.5-travis
- `clojure-mode-syntax-table/characters, clojure-mode-syntax-table/comment-macros` fail in every 4 travis builds, however they pass correctly on my machine (tested in 25.2 and also 25.3). Could it be that we have a glitch in the travis somewhere? Could you try to test it on your machine please?

> @Bost How is all of this shaping up?

In general, I've been using this branch for 2+ weeks w/o any problem. I think the RC1 is done.

Bost:
... I just stumbled upon an interesting article about https://en.wikipedia.org/wiki/Hibernation, hmm.

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@
 * [#429](https://github.com/clojure-emacs/clojure-mode/issues/429): Fix a bug causing last occurrence of expression sometimes is not replaced when using `move-to-let`.
 * [#423](https://github.com/clojure-emacs/clojure-mode/issues/423): Make `clojure-match-next-def` more robust against zero-arity def-like forms.
 * [#451](https://github.com/clojure-emacs/clojure-mode/issues/451): Make project root directory calculation customized by `clojure-project-root-function`.
+* Numerous font-locking fixes and improvements
```
bbatsov: List here the biggest user-visible changes explicitly and I guess we can merge this.

bbatsov:
> clojure-find-ns-test fails only in emacs-24.4-travis, emacs-24.5-travis

Does this work for you? I'm assuming the failure must be related to changing the ns regexp or something. 

> I (partially) fixed one of them, so:

Define partially. :-)

As for documenting it - I'd stick to high-level. Something like:

* stop font-locking interop methods, because of this and that
* stop font-locking constants because of this and that
* change ns font-locking in this than manner, etc

You can also post here some screenshot of before and after for people who'll stumble upon this ticket.

And sorry about the slow responses, but I've been crazy busy lately and I spend my little spare time focusing on some long overdue CIDER improvements.

Bost:
Examples of valid namespace definitions: before / after (i.e. with and without this change):
before
after

bbatsov:
Yeah, the ns part certainly looks good. I was also referring to the other things affected by this - e.g. constants and interop methods.

Anyways, update the changelog to be more granular and I guess we can merge this and gather more feedback for the users.

bbatsov:
@Bost ping :-)

Bost:
> As for documenting it - I'd stick to high-level. Something like:

I've done it. See the CHANGELOG.md

> > I (partially) fixed one of them, so:
> 
> Define partially. :-)

Oh BTW I was able to fix the failing emacs-25.1-travis and emacs-25.2-travis builds. The failing `clojure-mode-syntax-table/characters` - that was my mistake but the other one `clojure-mode-syntax-table/comment-macros`:
```
(should (equal (clojure-test-face-at
                  1 2  "#_ \n;; some crap\n (lala 0101\n lao\n\n 0 0i)")
                  'default))
```
that was IMO faulty. The '#_' char sequence wasn't matched by any font-locking regex so then its face should be `default` instead of `nil`.

Another thing: I wonder how was my predecessor able to pass the travis tests without ant long-line warning? See https://travis-ci.org/clojure-emacs/clojure-mode/jobs/352241172 @alexander-yakushev ?

bbatsov:
> Another thing: I wonder how was my predecessor able to pass the travis tests without ant long-line warning? See https://travis-ci.org/clojure-emacs/clojure-mode/jobs/352241172 @alexander-yakushev ?

Just another odd Travis problem I guess. 

Thanks for tackling this!

vspinu:
This shaved the highlighting of type hints. I find it much harder to read inter-op code now because types are easily confused with arguments. Any objection of putting `font-lock-type-face` back on type hints?

vspinu:
Actually the PR has provisioned type face for hints, but they don't work for me starting with this PR.

bbatsov:
@vspinu I can confirm this. I guess you can file another ticket so we won't forget to fix this.

Bost:
> @vspinu I can confirm this. I guess you can file another ticket so we won't forget to fix this.

Could you show some screenshots please? Or even better add a new test case. I haven't done much SW development in the last X weeks. I need to update myself at first. Thanks.

vspinu:
The type hints were fixed in dd8a193178.

PR #465, opened 2017-12-18
don't find namespace in Org-mode buffer for ob-clojure `cider-current…

stardiviner:
Because an org-mode file might contains multiple babel src blocks which has different Clojure namespace source code. If `clojure-find-ns` in an Org-mode will cause `ob-clojure` evaluate src block error "namespace-not-found" in cider eval respond dict.

For example:
```org
* test results output

#+BEGIN_SRC clojure :result output
(println "hi")
(println (str *ns*))
#+END_SRC

When I execute first src block [C-c C-c], it will find namespace and
return wrong namespace ~kk~ in second src block. This is not a
expected behavior.

* different namespace

#+BEGIN_SRC clojure :result output
(in-ns 'kk)
(println (str *ns*))
#+END_SRC
```
-----------------

Thanks!

### DIFF
--- a/clojure-mode.el
+++ b/clojure-mode.el
@@ -1761,19 +1761,20 @@ content) are considered part of the preceding sexp."
   "Return the namespace of the current Clojure buffer.
 Return the namespace closest to point and above it.  If there are
 no namespaces above point, return the first one in the buffer."
-  (save-excursion
-    (save-restriction
-      (widen)
+  (if (not (equal major-mode 'org-mode)) ; don't find namespace in Org-mode buffer for ob-clojure `cider-current-ns'.
+      (save-excursion
+        (save-restriction
+          (widen)
 
-      ;; Move to top-level to avoid searching from inside ns
-      (ignore-errors (while t (up-list nil t t)))
+          ;; Move to top-level to avoid searching from inside ns
+          (ignore-errors (while t (up-list nil t t)))
 
-      ;; The closest ns form above point.
-      (when (or (re-search-backward clojure-namespace-name-regex nil t)
-                ;; Or any form at all.
-                (and (goto-char (point-min))
-                     (re-search-forward clojure-namespace-name-regex nil t)))
-        (match-string-no-properties 4)))))
+          ;; The closest ns form above point.
+          (when (or (re-search-backward clojure-namespace-name-regex nil t)
+                    ;; Or any form at all.
+                    (and (goto-char (point-min))
+                         (re-search-forward clojure-namespace-name-regex nil t)))
+            (match-string-no-properties 4))))))
 
 (defconst clojure-def-type-and-name-regex
   (concat "(\\(?:\\(?:\\sw\\|\\s_\\)+/\\)?"


clojure-mode.el
```diff
@@ -1761,19 +1761,20 @@ content) are considered part of the preceding sexp."
...
-      ;; Move to top-level to avoid searching from inside ns
-      (ignore-errors (while t (up-list nil t t)))
-
-      ;; The closest ns form above point.
-      (when (or (re-search-backward clojure-namespace-name-regex nil t)
-                ;; Or any form at all.
-                (and (goto-char (point-min))
-                     (re-search-forward clojure-namespace-name-regex nil t)))
-        (match-string-no-properties 4)))))
+  (if (not (equal major-mode 'org-mode)) ; don't find namespace in Org-mode buffer for ob-clojure `cider-current-ns'.
```
bbatsov: This should be a change in CIDER, not in `clojure-mode`.

stardiviner: Thanks for your hint, I found another solution.
I checked out `cider-buffer-ns` in `cider-current-ns`.
Then come up with this solution:

          #+begin_src emacs-lisp
          ;; auto start CIDER REPL session in a complete Leiningen project environment for Org-mode Babel by jack-in.
          (add-to-list 'org-babel-default-header-args:clojure
                       '(:session . "*cider-repl ob-clojure*"))
          (progn
            (find-file (expand-file-name "~/.emacs.d/Org-mode/ob-clojure/src/ob_clojure/core.clj"))
            (cider-jack-in))

          (defun ob-clojure-cider-do-not-find-ns ()
            "Fix the issue that `cider-current-ns' try to invoke `clojure-find-ns' to extract ns from buffer."
            (setq-local cider-buffer-ns "user"))
          (add-hook 'org-mode-hook #'ob-clojure-cider-do-not-find-ns)
          #+end_src

Thanks @bbatsov

stardiviner: Hope this solution can help someone want it.

stardiviner: I got an better function:
```elisp
(defun ob-clojure-cider-do-not-find-ns ()
  "Fix the issue that `cider-current-ns' try to invoke `clojure-find-ns' to extract ns from buffer."
  (with-current-buffer "*cider-repl ob-clojure*"
    (defvar ob-clojure-cider-repl-ns cider-buffer-ns)
    (setq-local cider-buffer-ns ob-clojuure-cider-repl-ns)))
```
This makes it smarter.

bzg: @bbatsov If a change is possible directly in CIDER it would be better than the solution proposed here I guess.  I'm AFK for a week but if you can give us some directions on what to change, I will have a look. Thanks!

bbatsov: Yeah, of course. Basically this major-mode check should be in `cider-find-ns`, as it naturally assumes it's working in a `clojure-mode` buffer. That's a trivial change.

The other thing is to decide whether it's ok to just assume in `org-mode` the ns is always going to be to `user`, the current REPL's namespace, or something else.

stardiviner: I agree @bbatsov 's thinking. So I come up with this temp workaround. After @bbatsov 's mention, I still need to improve my code to always dynamic update ns for `ob-clojure-literate`.

mvarela: I just came across this problem this week, and the problem is that you could have several namespaces in a single org-mode file (depending on how you organize your literate sources), so just assuming any default option is bound to be wrong in some cases. From the ones mentioned by @bbatsov , probably the current REPL's ns would make the most sense.

In the end, I found another workaround, extending the header options for Clojure blocks in org-babel via a `:before` advice. Details in this gist: https://gist.github.com/mvarela/08663bafd54c60d478b26b69a207453a

bzg: Would something like the patch below work for you? I cannot test it myself now, can you do so and report problems? Thanks!

[ob-clojure.el.patch.txt](https://github.com/clojure-emacs/clojure-mode/files/1831822/ob-clojure.el.patch.txt)

stardiviner: After test, I think this patch will be great. Thanks. Will it be in available in Org-mode soon?

stardiviner: And because `cider-jack-in` outside of project by default namespace is `user`, so can let `ob-clojure` `ns` be a namespace retrived from available CIDER REPL connection? or by default with `user`?

bbatsov: So, basically this sets the `cider-buffer-ns` each time you evaluate something in a block? What should be the fallback if no `:ns` metadata exists? Also I wonder if it wouldn't be easier to just pass the extracted ns directly to some evaluation command (I'm not sure which one exactly are you using there, instead of setting the buffer-local).

stardiviner: Sorry, @bbatsov  I make a mistake on names. I should notify @bzg . Ahuuu... I think his patch should works good. I will email him.

bbatsov: No problem. As mentioned in my comment I think that solution can probably be improved upon by passing the ns directly to each evaluation, which would make it much more transparent what exactly are you doing, but this should probably work.

stardiviner: You mean use `(in-ns <namespace>)`? That will need to add a clojure line. I see `org-babel-expand-body:clojure`. It should can include this. I misunderstand your meaning about that. I always have this communication trouble on English. suggestion accepted.

bbatsov: I mean like passing this `:ns` you're extracting from the header directly to whatever cider evaluation command you're actually using internally in `ob-clojure`. Not sure why doesn't hasn't been done like this to begin with.

bbatsov: The conversation so far leads me to believe you're using directly something like `cider-eval-last-sexp`, but if I were you I'd add some ob-clojure specific evaluation handler that just gets the sexp and the ns from the org buffer and passes those for evalution to one of cider's lower-level commands (e.g. cider-interactive-eval) or something like this.

stardiviner: No, I have not use `cider-eval-last-sexp` or similar thing.
You mean like this? https://gist.github.com/480fc2552161473caab589bc8a8a0f60

stardiviner: This patch comes a problem, it does change namespace, but it is permanently, if the second src block have not specified `:ns`, the second src block still in `:ns` specified namespace instead of CIDER REPL default namespace. Then user will have to specify `:ns` for every src block to avoid confusing.

stardiviner: I pasted src block examples and result in [gist comment](https://gist.github.com/stardiviner/480fc2552161473caab589bc8a8a0f60#gistcomment-2391856)

bbatsov: Yeah, that's why I felt this wasn't a good solution. I guess a simple workaround would be to set this variable to either `user` or the current REPL's ns, if there's no :ns header, but that instead of this dance I stand by my suggestion to just pass an ns directly to the eval command of present, otherwise fallback to something else there. Btw, if you use a low-level eval command this ns inference won't be a problem for you at all as this was obviously meant for clojure-mode and cider-scratch-mode buffers, which have different semantics.

stardiviner: You said a low-level eval command, which kind of they are? `nrepl-*` prefix like? In ob-clojure function `org-babel-execute:clojure`, it use `nrepl-sync-request:eval`.

bbatsov: So, you've got your solution then - see `nrepl-sync-request:eval (input connection &optional ns tooling)`. Instead of using something like `clojure-find-ns` or `cider-find-ns`, which you shouldn't be doing at all, you should just extract this :ns header that was proposed and pass it to this function. You just need to decide what fallback ns to use in case there's no `:ns` specified and problem solved.

stardiviner: I found the cause, In `org-babel-execute:clojure`, `nrepl-sync-request:eval` use `(cider-current-ns)` as parameter. That why caused upper problem. And if `:ns` does not exist in Clojure REPL, it will return error namespace does not exist. I think need to send `:ns` eval before send the body. But this will make it more complex. I will take a try. coding.

mvarela:
I think there are different use cases to consider when doing Clojure from
org-mode, which are different than when you're editing a Clojure namespace
directly in a .clj file (this comes from how I use it, but it is not
exhaustive).
First, it is common to use org files as a sort of "notebook" (à la
Jupyter). For this type of interactive use, being in the `user` NS is not
necessarily a problem (but it might be good to have alternatives, if we
want to tangle that code later on).
Then, another use case is to do literate programming. In this case, it may
well be that there are several Clojure NS in a single org-mode file, and no
default will be likely to cover all situations. When I proposed the
adivce-based solution (adding the :ns header to ob-clojure blocks), my idea
is that I can add that after declaring my NS, and subsequent src blocks in
the org-mode buffer will be evaluated in that NS until I change it again.
This assumes that you're laying out your code in a more or less sequential
manner inside the org buffer, but it's not a huge assumption (at least for
my workflow, it holds pretty good).

stardiviner:
@mvarela your proposed advice solution I already done in `ob-clojure-literate` at [here](https://code.orgmode.org/bzg/org-mode/src/master/contrib/lisp/ob-clojure-literate.el#L150)
I think use `:ns` is better solution. I forget the tangling case. It need to tangle `(ns <namespace>)` clojure code too. I my new implement, it should be included and handled too.
Here is a new [patch](https://gist.github.com/8de240ea523a3fc00da2bad8c375c1a1). check it out.

stardiviner:
updated the patch a lot, here is the revision comparing. https://gist.github.com/stardiviner/8de240ea523a3fc00da2bad8c375c1a1/revisions

stardiviner:
I submitted PATCH to Org-mode and get merged now. Just as a remind.
https://code.orgmode.org/bzg/org-mode/commit/d7e12d1df7091563c5f0fe0bd8b2db634d3e87ba

bzg:
I will have a look, thanks for the heads up.

PR #466, opened 2018-01-01
Make three font-lock faces customizable

vemv:
Introduce `clojure-global-constant-face`, `clojure-def-symbol-face`, `clojure-lambda-arg-face`.

Updated tests, although I couldn't get to run them (never had done so. Tried with `M-x ert`):

```
Selector: t
Passed:  0
Failed:  0
Skipped: 0
Total:   0/0

Started at:   2018-01-01 07:14:57+0100
Finished.
Finished at:  2018-01-01 07:14:57+0100
```

### DIFF
--- a/clojure-mode.el
+++ b/clojure-mode.el
@@ -92,6 +92,26 @@
   "Face used to font-lock Clojure character literals."
   :package-version '(clojure-mode . "3.0.0"))
 
+(defface clojure-interop-method-face
+  '((t (:inherit font-lock-preprocessor-face)))
+  "Face used to font-lock interop method names (camelCase)."
+  :package-version '(clojure-mode . "3.0.0"))
+
+(defface clojure-global-constant-face
+  '((t (:inherit font-lock-constant-face)))
+  "Face used to font-lock the Clojure global constants: nil, true, false."
+  :package-version '(clojure-mode . "3.0.0"))
+
+(defface clojure-def-symbol-face
+  '((t (:inherit font-lock-keyword-face)))
+  "Face used to font-lock the symbols `def`, `defn`, and all detected variants (be them from clojure.core or not)."
+  :package-version '(clojure-mode . "3.0.0"))
+
+(defface clojure-lambda-arg-face
+  '((t (:inherit font-lock-variable-name-face)))
+  "Face used to font-lock the function literal args: %, %&, %1 and so on."
+  :package-version '(clojure-mode . "3.0.0"))
+
 (defcustom clojure-indent-style :always-align
   "Indentation style to use for function forms and macro forms.
 There are two cases of interest configured by this variable.
@@ -754,7 +774,7 @@ any number of matches of `clojure--sym-forbidden-rest-chars'."))
                 ;; Possibly type or metadata
                 "\\(?:#?^\\(?:{[^}]*}\\|\\sw+\\)[ \r\n\t]*\\)*"
                 "\\(\\sw+\\)?")
-       (1 font-lock-keyword-face)
+       (1 'clojure-def-symbol-face)
        (2 font-lock-variable-name-face nil t))
       ;; Type definition
       (,(concat "(\\(?:clojure.core/\\)?\\("
@@ -767,7 +787,7 @@ any number of matches of `clojure--sym-forbidden-rest-chars'."))
                 ;; Possibly type or metadata
                 "\\(?:#?^\\(?:{[^}]*}\\|\\sw+\\)[ \r\n\t]*\\)*"
                 "\\(\\sw+\\)?")
-       (1 font-lock-keyword-face)
+       (1 'clojure-def-symbol-face)
        (2 font-lock-type-face nil t))
       ;; Function definition (anything that starts with def and is not
       ;; listed above)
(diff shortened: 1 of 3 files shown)


bbatsov:
Two notes from me:

* This overlaps with https://github.com/clojure-emacs/clojure-mode/pull/462 which suggest dropping constant font-locking for everything but those 3 symbols
* I don't think that the term for them is really global constants, although I'm not sure what is it, technically speaking

vemv:
* Broadened the scope of this PR to span a couple more deffaces I needed
* Can rename to `boolean-or-nil`, sounds good?
* Regarding #462, it doesn't seem all that related (although I can wait for it to be merged; would avoid obvious merge conflict). In that PR, two `defface`s still exist. I think the more deffaces the better, else syntactically unrelated tokens get a coupled coloring.

IOW, clojure-mode currently decides that `def` and `cond` must get the same coloring, which can be frustrating. The more this is parameterized the better?

(I could open a PR introducing 12 deffaces... but better to go on an as-needed basis)

bbatsov:
> Regarding #462, it doesn't seem all that related (although I can wait for it to be merged; would avoid obvious merge conflict).

image

clojure-mode.el
```diff
@@ -97,6 +97,21 @@
   "Face used to font-lock interop method names (camelCase)."
   :package-version '(clojure-mode . "3.0.0"))
 
+(defface clojure-global-constant-face
+  '((t (:inherit font-lock-constant-face)))
+  "Face used to font-lock the Clojure global constants: nil, true, false."
+  :package-version '(clojure-mode . "3.0.0"))
+
+(defface clojure-def-symbol-face
```
bbatsov: I'm don't like this idea. There's nothing special about definitions that warrants a different face for them.

clojure-mode.el
```diff
@@ -97,6 +97,21 @@
...
+  '((t (:inherit font-lock-keyword-face)))
+  "Face used to font-lock the symbols `def`, `defn`, and all detected variants (be them from clojure.core or not)."
+  :package-version '(clojure-mode . "3.0.0"))
+
+(defface clojure-lambda-arg-face
```
bbatsov: I don't understand why they can't be font-locked as variables, as this is what they are. That seems like an odd decision to me. And I don't think we use the variable face for anything but them.

bbatsov:
> else syntactically unrelated tokens get a coupled coloring.

But what you're suggesting goes against the spirit of what your saying - we don't care about definitions, we care about `macros`, `special forms`, etc. I can imagine someone wanted to be able to differentiate a special form form a macro, but Emacs modes have traditionally avoided this, so I just followed suit here. Having different faces for the same syntactic category (e.g. `defn` and `cond`) seems both confusing and excessive. 

I general I'm ok to add some extra faces for things that are not part of the standard Emacs font-locking, but I'm not ok with adding faces for subjective reasons.

vemv:
Hi there, thanks for sharing your reasoning!

> but Emacs modes have traditionally avoided this, so I just followed suit here

Fair enough, although it also can be frustrating to be limited by design decisions made maybe decades ago.

font-lock presumes that all languages' syntaxes can be mapped 1-1 to each other, e.g. all languages have "constants" (therefore `font-lock-constant-face` exists).

That's a noble goal to seek, and probably it's good enough 95% of the times. But eventually someone will want finer-grained control for his favorite language.

---

Maybe if I explain my exact use case you'll see it better. Corrections welcome.

* I like core macros / special forms with a distinct coloring -plain yellow-, helps me see the flow of my code
* I like `foo` (in `def foo`) to be prominently colored -darker yellow-, so all those names are skimmable in a given ns:

image

It's great that those names are in a distinct, stronger color from anything else. That way my eyes can easily find the `def` I'm looking for.

* But: `def` is a special form, and `foo` (from `def foo`) is intended to be prominent. See what happens without my patch applied:

image

The names aren't prominent anymore. So, the symbols `def/defn` should be as plain as possible, like in my previous screenshot. But I cannot do that without affecting `cond`, `if` etc.

---

> I general I'm ok to add some extra faces for things that are not part of the standard Emacs font-locking, but I'm not ok with adding faces for subjective reasons.

I hope I gave a good-enough rationale! In any case, we can make the docstrings advocate the "normal" thing:

```diff
- "Face used to font-lock the Clojure global constants: nil, true, false."
+ "Face used to font-lock the Clojure global constants: nil, true, false. Note that you should normally configure font-lock-constant-face instead"
```

WDYT?

Cheers - Victor

bbatsov:
Seems to me a part of your problem is some odd decision in your color-theme to use very similar colors for special forms and function names. In mine this looks like this:

image

> WDYT?

I'll ruminate on the subject.

vemv:
Rumination appreciated!

I see the point of your screenshot, it's nice but if you look at the third defn, there's no contrast between `defn` and `case` it, it looks "chunky".

Generally I strive to keep things visually separated: `def` from `<name>`, declaration "headers" from body, and so on. That way I can parse things more quickly.

vemv:
Updated PR to solve merge conflicts.

Could you give a think to the pending disagreement?

bbatsov:
I completely forgot about this. The first face doesn't really make sense anymore as we dropped the notion of special face for interop forms. We also dropped the notion of a special face for SNAKE_CASE constants, I guess having a special face for 3 constants is pointless now. 

And I still don't think that definitions warrant a special face, so I'm inclined to just reject this suggestion. @clojure-emacs/cider any thoughts from anyone else?

vemv:
> I completely forgot about this.

No issue, thanks for the reply!

I was thinking, while my goal is somewhat reasonable (namely: to color Clojure as I deem semantic), the needs of 1 developer shouldn't impact the clojure-mode codebase.

Perhaps a cleaner solution would be to separate the relevate .el parts to a different `.el` file, so I can cleanly, fine-grainedly replace it (instead of maintaining a personal clojure-mode fork, as I do right now).

So basically you'd have the same code as today, but split into more files.

bbatsov:
Well, you can always replicate the approach of https://github.com/clojure-emacs/clojure-mode/blob/master/clojure-mode-extra-font-locking.el

This is some font-locking that used to live in clojure-mode, but I deemed it too unreliable and removed it, but I still provided it as a package in case someone really liked it.

PR #477, opened 2018-06-02
Add shadow-cljs.edn to the default clojure build tool files list

cdorrat:
Cider now supports shadow-cljs builds, this commit adds it's build file (shadow-cljs.edn) to the default list of build tool files so that ```clojure-project-dir``` works as expected for these projects without additional configuration.


bbatsov:
👍

PR #484, opened 2018-08-12
Make `beginning-of-defun` aware of clojure comment form

dpsutton:
Moves the notion of top level inside of a comment form into `clojure-mode`. Makes the necessary change such that `beginning-of-defun` is aware of this and thus CIDER can use it transparently.

Added tests and changelog. Wasn't sure if this should be in README. 

Next step will be to pull this change out of CIDER once this is accepted.

### DIFF
--- a/clojure-mode.el
+++ b/clojure-mode.el
@@ -531,7 +531,8 @@ replacement for `cljr-expand-let`."
   (setq-local clojure-expected-ns-function #'clojure-expected-ns)
   (setq-local parse-sexp-ignore-comments t)
   (setq-local prettify-symbols-alist clojure--prettify-symbols-alist)
-  (setq-local open-paren-in-column-0-is-defun-start nil))
+  (setq-local open-paren-in-column-0-is-defun-start nil)
+  (setq-local beginning-of-defun-function #'clojure-beginning-of-defun-function))
 
 (defsubst clojure-in-docstring-p ()
   "Check whether point is in a docstring."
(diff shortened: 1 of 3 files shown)


dpsutton:
tests fail because of `if-let*`. i thought that was the preferred way

clojure-mode.el
```diff
@@ -555,7 +555,9 @@ replacement for `cljr-expand-let`."
   (add-hook 'electric-indent-functions
             (lambda (_char) (if (clojure-in-docstring-p) 'do-indent)))
   ;; integration with project.el
-  (add-hook 'project-find-functions #'clojure-current-project))
+  (add-hook 'project-find-functions #'clojure-current-project)
+  (set (make-local-variable 'beginning-of-defun-function)
```
bbatsov: All this setup is now in `clojure-mode-variables`.

clojure-mode.el
```diff
@@ -1943,6 +1946,65 @@ This will skip over sexps that don't represent objects, so that ^hints and
           (backward-sexp 1))
         (setq n (1- n))))))
 
+(defun clojure-top-level-comment-p ()
```
bbatsov: I think we should make this more generic by making it take `comment` or whatever other form as a param.

bbatsov:
@dpsutton I see a failure because of  `seq-find`.

clojure-mode.el
dpsutton: This would just return the first form of a sexp? or compare the first form to the arg passed in?

dpsutton:
@bbatsov re: `seq-find`. I can replace this but `seq-find` is just fine in CIDER. are you against adding a dep on cl-seq?

bbatsov:
> seq-find. I can replace this but seq-find is just fine in CIDER. are you against adding a dep on cl-seq?

I guess you mean `seq`, right? :-) No, it's a built-in in Emacs 25+, so that's fine by me. We didn't use it before simply for that reason. Generally, however, it's best for the code to be consistent and use only `cl-lib` functions or their counterparties in `seq`. The tests were failing simply because `seq` wasn't required anyway. I assume `cl-lib` should have similar function. 

Now it seems the build is failing due to a missing docstring.

bbatsov:
👍

PR #487, opened 2018-08-24
Fix bug in end-of-defun

dpsutton:
Going to end of defun would skip over two forms. Needed to pass on the negative
prefix that was used by the generic parts of this mechanism.

Also fixed an issue with paredit would not insert parens in an empty
buffer by handling the `end-of-buffer` condition.

### DIFF
--- a/clojure-mode.el
+++ b/clojure-mode.el
@@ -1955,7 +1955,8 @@ This will skip over sexps that don't represent objects, so that ^hints and
         (clojure-forward-logical-sexp 1)
         (clojure-backward-logical-sexp 1)
         (looking-at-p first-form))
-    (scan-error nil)))
+    (scan-error nil)
+    (end-of-buffer nil)))
 
 (defun clojure-sexp-starts-until-position (position)
   "Return the starting points for forms before POSITION.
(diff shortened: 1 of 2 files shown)


clojure-mode.el
```diff
@@ -1992,10 +1993,11 @@ testing, give an easy way to turn this new behavior off."
         (setq haystack (cdr haystack))))
     found))
 
-(defun clojure-beginning-of-defun-function ()
+(defun clojure-beginning-of-defun-function (&optional n)
   "Go to top level form.
 Set as `beginning-of-defun-function' so that these generic
-operators can be used."
+operators can be used. Given a positive N it will do it that many
```
bbatsov: I think the lack of double space after `.` is failing the build.

dpsutton: haha i bet you are correct. one second

test/clojure-mode-sexp-test.el
```diff
@@ -65,7 +65,21 @@ and point left there."
...
+
+(ert-deftest test-clojure-end-of-defun-function ()
+  (clojure-buffer-with-text
+      "
+(first form)
+|
+(second form)
+
+(third form)"
+
```
bbatsov: The code after the string seems misindented.

bbatsov:
👍

PR #488, opened 2018-09-02
Introduce `clojure-align-separator` defcustom

### LINKED ISSUE #481: Blank lines prevent correct vertical alignment
Assuming I have vertical alignment turned on for maps, the following won't get aligned due to the blank line:

```clojure
{:keys [foo bar

        baz]
 :or {quux 1}}
```

Tried with `clojure-mode-20180809.1455`.

Thanks - Victor

vemv:
Fixes #481 

I'm not sure I used the `type '(choice (const ...` syntax correctly, since when I changed it using the value menu in `M-x customize variable`, the value of `clojure-align-separator` would appear unchanged  (even though the widget says otherwise).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### New features
 
+* [#481](https://github.com/clojure-emacs/clojure-mode/issues/481): Support vertical alignment even in the presence of blank lines, with the new `clojure-align-separator` user option.
 * [#483](https://github.com/clojure-emacs/clojure-mode/issues/483): Support alignment for reader conditionals, with the new `clojure-align-reader-conditionals` user option.
 
 ## 5.9.1 (2018-08-27)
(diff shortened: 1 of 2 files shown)


Malabarba:
There's one quote too-many in the choice format.
The way you've defined it, the first value is the symbol `clojure--align-separator-newline-regexp`, and the second value is the list `(quote entire)`.
I think you need to replace it with this:

```emacs-lisp
`(choice (const :tag "Make blank lines prevent vertical alignment from happening."
                ,clojure--align-separator-newline-regexp)
         (other :tag "Allow blank lines to happen within a vertically-aligned expression."
                entire))
```

vemv:
Thanks for the help! Tried out, works correctly. Commit amended.

`:safe` and `:type` values are missing because I couldn't find out the syntax for an option between symbol and regexp.

vemv:
Status?

bbatsov:
Ops, sorry about that. It completely slipped my mind. Thanks!

Issue #489, opened 2018-09-03
clojure-toplevel-inside-comment-form broke parens insertion before comment block.


DogLooksGood:
### Actual behavior and reproduce steps

with `clojure-toplevel-inside-comment-form` set, and `paredit` enabled. Insert a pair of parens
```
()

|   ;; <-- cursor here, type `(`

(comment)
```
will become 
```
()



(|)(comment)  ;; <-- cursor jump to here
```

### clojure-mode version information

```
clojure-mode (version 5.8.2)
```

### Emacs version

GNU Emacs 26.1 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.22.30) of 2018-07-05

### Operating system

Linux

bbatsov:
Sounds like a task for @dpsutton. :-)

dpsutton:
This is weird interaction wit paredit mode but i'm investigating.

PR #490, opened 2018-09-23
[Fix 489] Inserting parens before comment form doesn't move point

dpsutton:
https://github.com/clojure-emacs/clojure-mode/issues/489

In a form like

|

(comment
   (stuff))

Entering parens with paredit would put the parens right before the comment
block. Paredit determines if it is in a comment to insert parens so it doesn't
automatically enter a closing when in a comment or a string. Part of this called
beginning-of-defun which we have modified.

The error here was that rater than just going to the beginning of the form, we
went to the end and then back one logical form to be at the beginning. This is
identical behavior _unless_ you are between two forms. Going straight to the
beginning put you in the first form, going to the end and then the beginning
puts you in the second form. I.e.,

(formA)
|
(formB)

Our beginning of form went to formB but it should go to formA.

-----------------

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 * [#483](https://github.com/clojure-emacs/clojure-mode/issues/483): Support alignment for reader conditionals, with the new `clojure-align-reader-conditionals` user option.
 
+### Bugs fixed
+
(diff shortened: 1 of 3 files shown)


dpsutton:
@bbatsov ping :)

PR #491, opened 2018-09-27
Fix typo

mikepjb:
This is a typo fix in a comment, it should not affect the behaviour of clojure-mode. It's a small thanks for creating this mode!

I ran this using emacs version 26.0.50.3 on OSX El Capitan, the clojure-mode version is shown as 5.8.2. I don't believe there are any other tasks to take care of as described in the contribution guidelines.


bbatsov:
Thanks!

PR #495, opened 2018-10-24
Fix font-lock of type hints

vspinu:
Currently type hints are matched as starting with `#^foo` but I think most users use the plain `^boo`. I personally didn't even know that type hinting can be done with `#^`.

### DIFF
--- a/clojure-mode.el
+++ b/clojure-mode.el
@@ -439,7 +439,7 @@ ENDP and DELIM."
               (t)))))
 
 (defconst clojure--collection-tag-regexp "#\\(::[a-zA-Z0-9._-]*\\|:?\\([a-zA-Z0-9._-]+/\\)?[a-zA-Z0-9._-]+\\)"
-    "Collection reader macro tag regexp.
+  "Collection reader macro tag regexp.
 It is intended to check for allowed strings that can come before a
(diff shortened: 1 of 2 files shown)


bbatsov:
Are you sure the change won't affect something like `(def ^:private`? I also wonder if we shouldn't just font-lock the entire reader macro + the actual type. I think that will stand out more.

vspinu:
> Are you sure the change won't affect something like (def ^:private? 

It does not. 

> I also wonder if we shouldn't just font-lock the entire reader macro + the actual type. I think that will stand out more.

Do you mean to highlight `^` in front? I tried and didn't like it. One would need an extra cognitive effort to distinguish the ^ from the ordinary letter. Right now it's very clear that it's a type hint. That perk of different color plays a brain trick of some sort.

bbatsov:
Ah, seems the tests mislead me. I was just puzzled why we font-lock some part of type with the type face, and the other with no face. Seems the `/` is messing up things, but I guess in practice for types that's rarely going to be an issue.

vspinu:
The highlighting of the thing after "/" is set to 'default by design. If you meant that the second word after / should be highlighted in the type face then yes, it would make some sense. I guess the current situation is for the sake of compatibility with namespaced symbols which are highlighted in mixed color.

PR #497, opened 2018-11-17
Indent "let", "when" and "while" as function form if not at start

Macroz:
"let", "when" and "while" are considered to be a macro form normally:
```
(when foo
  bar)
```
Also you can introduce macros that start with "let", "when" or "while" that
will be indented like macro forms:
```
(when-foo-bar
  1
  2
  3)
```
But when "let", "when" and "while" are not in the beginning of a symbol they are
now threated as function forms:
```
(foo-when-bar 1
              2
              3)
```
-----------------

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 * [#481](https://github.com/clojure-emacs/clojure-mode/issues/481): Support vertical alignment even in the presence of blank lines, with the new `clojure-align-separator` user option.
 * [#483](https://github.com/clojure-emacs/clojure-mode/issues/483): Support alignment for reader conditionals, with the new `clojure-align-reader-conditionals` user option.
-- Indent "let", "when" and "while" as function form if not at start of a symbol
+* [#497](https://github.com/clojure-emacs/clojure-mode/pull/497) Indent "let", "when" and "while" as function form if not at start of a symbol
 
 ### Bugs fixed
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 * [#481](https://github.com/clojure-emacs/clojure-mode/issues/481): Support vertical alignment even in the presence of blank lines, with the new `clojure-align-separator` user option.
 * [#483](https://github.com/clojure-emacs/clojure-mode/issues/483): Support alignment for reader conditionals, with the new `clojure-align-reader-conditionals` user option.
+- Indent "let", "when" and "while" as function form if not at start of a symbol
```
bbatsov: Please, make this entry consistent with the other entires in the changelog (e.g. link PR, use `*` instead of `-`, add `.`).

clojure-mode.el
```diff
@@ -1304,7 +1304,7 @@ symbol properties."
                  'clojure-indent-function)
             (get (intern-soft (match-string 1 function-name))
                  'clojure-backtracking-indent)))
-      (when (string-match (rx (or "let" "when" "while") (syntax symbol))
+      (when (string-match (rx string-start (or "let" "when" "while") (syntax symbol))
```
bbatsov: I guess it's also worth adding some explanations in the code here, as to what it's aiming to achieve. You should also check if this won't break `if-let`.

CHANGELOG.md
Macroz: Thanks, fixed the chicken-and-egg plus wrong list item type

clojure-mode.el
Macroz: Are the new explanations something you had in mind?

I also added tests for if-let case and fixed it.

bbatsov:
Thanks!

PR #498, opened 2018-11-17
Describe common development setup for contributors

Macroz:
Describe common development setup for contributors. 

I got help from the Slack and thought that it would be nice to have these here.

I'm happy to improve the documentation more if you have suggestions.

-----------------

Thanks!


bbatsov:
macOS users should also make sure that `emacs` points to whatever Emacs they installed as opposed to the built-in Emacs 22. :-)

PR #500, opened 2018-12-06
Add project-roots method for project.el compatibility

vspinu:
For full compatibility with project.el one needs to define a `project-roots` method. (emacs-ess/ESS#786)


bbatsov:
Thanks!

PR #503, opened 2019-01-05
Fix Makefile so that we can compile again

arichiardi:
The functionality has been ported from cider, we can now use the make command
to install elpa packages and compile. This is still not perfect because in
theory we should not install things from elpa if we have them in our Emacs but
it would take more work for achieving that.

-----------------

Thanks!


cichli:
Thanks!

Issue #469, opened 2018-02-06
Unexpected indentation of top-level string-valued def


codeasone:
Perhaps related to https://github.com/clojure-emacs/clojure-mode/issues/405 

## Expected behavior

```
(def foo "<character-entered>abc")
``` 

## Actual behavior

```
  (def foo "<character-entered>abc")
``` 

Form indented as a docstring unnecessarily. 

## Steps to reproduce the problem

With the following `var` definition: 
```
(def foo "|abc")
``` 
And cursor at `|` 

Entering any character causes unhelpful indentation:

```
  (def foo "Mabc")
``` 

It appears to be because `clojure-mode` determines that the form is a [docstring](https://github.com/clojure-emacs/clojure-mode/blob/5cf0fd9360dc5a9a95464601319062673d213807/clojure-mode.el#L1207)

Unfortunately I don't have the time to debug the treatment of [syntax-ppss](https://www.gnu.org/software/emacs/manual/html_node/elisp/Position-Parse.html), which I suspect is at the root of the issue. 

## Environment & Version information

```
GNU Emacs 26.0.91 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.18.9) of 2018-01-26
clojure-mode (version 5.7.0-snapshot)
Version: 20180202.922
Commit: 5cf0fd9360dc5a9a95464601319062673d213807
```

The issues does not present with current `melpa-stable` release of `clojure-mode`: 

```
Version: 5.6.1
Commit: fdbdfb91e0a5731bf9a739b70c5f217c5d12ae6d
```

bbatsov:
Yeah, that's basically the same bug. I have a good idea where the problematic code is, but I don't have time to debug it myself either. Adding a real docstring should work as a workaround I guess. 

It's basically a combination of

https://github.com/clojure-emacs/clojure-mode/blob/master/clojure-mode.el#L906

and

https://github.com/clojure-emacs/clojure-mode/blob/master/clojure-mode.el#L1035

We mark the docstring positional statically and we probably need to make this more dynamic homehow (e.g. it's number 2 only if the form has 3 arguments).

cichli:
Thanks for the report.

Issue #405, opened 2016-08-26
Indentation bug involving strings 


sooheon:
## Expected behavior

Given the following with cursor at `|`:

```
(def foo |"abc")
```

`indent-for-tab-command` should make no change. The sexp is already correctly indented.
## Actual behavior

```
  (def foo |"abc")
```

Sexp is indented two spaces.
## Steps to reproduce the problem

The above case. When the string is on the next line, it is not reproduced.
### clojure-mode version information

```
clojure-mode (version 5.5.2)
```
### Emacs version

Emacs 25.1.1
### Operating system

OS X El Cap

Malabarba:
Verified.

drewverlee:
This happens for me on Ubuntu with spacemacs. Not sure what other information would be helpful.

codeasone:
I'm having a similar issue: 

```
GNU Emacs 26.0.91 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.18.9) of 2018-01-26
clojure-mode (version 5.7.0-snapshot)
Version: 20180202.922
Commit: 5cf0fd9360dc5a9a95464601319062673d213807
```

With the following `var` definition: 
```
(def foo "|abc")
``` 
And cursor at `|` 

Entering any character causes unhelpful (and annoying) indentation:

```
  (def foo "Mabc")
``` 

It appears to be because `clojure-mode` determines that the form is a [docstring](https://github.com/clojure-emacs/clojure-mode/blob/5cf0fd9360dc5a9a95464601319062673d213807/clojure-mode.el#L1207)

Unfortunately I don't have the time to debug the treatment of [syntax-ppss](https://www.gnu.org/software/emacs/manual/html_node/elisp/Position-Parse.html), which I suspect is at the root of the issue.

cichli:
Thanks for the report.

PR #505, opened 2019-01-20
[Fix #445] def form with strings and docstrings

### LINKED ISSUE #445: (def s "a string") font locks "a string" as a docstring
## Expected behavior

Font lock the string `"a string"` as a string in `(def s "a string")`.

## Actual behavior

The string `"a string"` is font locked as a docstring in `(def s "a string")`.

## clojure-mode version information
```
clojure-mode (version 5.7.0-snapshot)
```

## Proposed solution

I played around with this, and it can be solved by adding a check in this `and` https://github.com/clojure-emacs/clojure-mode/blob/master/clojure-mode.el#L898 in the function `clojure-font-lock-syntactic-face-function`. For example
```
(and ...
  (or (save-excursion
        (goto-char startpos)
        (not (= ?\) (char-after (+ startpos (length (sexp-at-point)) 2)))))
      (string= "defprotocol" firstsym)))
```

This, of course, checks for a closing paren after the string, and if so, it is not considered a docstring, except for `defprotocol`, which is the only kind of form that can have a docstring in the last position. Instead of checking for a paren, one could of course check for another element, but it is more involved.

Opinions, suggestions, etc?

carlosgeos:
I took the proposed solution here: https://github.com/clojure-emacs/clojure-mode/issues/445 (thanks @oskarkv) and scoped it to only the def form.

Basically:
```clojure
(def my_string_with_docstring "how to use" "just a string")
```
`how to use` and `just a string` are highlighted differently.

And:
```clojure
(def im_a_string "hello")
```
`hello` is no longer highlighted as a docstring

I did what was said in these comments: https://github.com/clojure-emacs/clojure-mode/issues/445#issuecomment-340689656 and https://github.com/clojure-emacs/clojure-mode/issues/445#issuecomment-341931740. Now it works for defs, defprotocol, ns (whatever the form)...

-----------------


bbatsov:
Please add some unit tests and update the change log as well. (This suggestions applies to the other PRs you’ve opened as well)

carlosgeos:
done for this one, the others will come soon

test/clojure-mode-indentation-test.el
```diff
@@ -68,9 +68,11 @@ values of customisable variables."
            (search-forward "|")
            (delete-char -1)
            (clojure-mode)
+           (font-lock-ensure)
            (indent-according-to-mode)
 
-           (should (equal expected-state (buffer-string)))
+           (should (equal expected-state (buffer-substring-no-properties
```
bbatsov: Why is this change necessary?

carlosgeos: Because `clojure-in-docstring-p` checks for a font-lock-face, `indent-according-to-mode` works differently (in this case) if font-lock-mode is off, so I wrote `font-lock-ensure` since otherwise there is no font-lock information in the temp-buffer.

If this change wasn't there, this test: 

https://github.com/clojure-emacs/clojure-mode/blob/43e3e65966b8798968016bf0134a726c4a73461e/test/clojure-mode-indentation-test.el#L123-L125

would always pass, both in the current version and my version. To make it fail in the current version (which is expected and wanted), the change had to be made. I found no other way to mimic what is really happening.

But you are right, I'll compare the strings without removing the properties, no reason to do that.

test/clojure-mode-indentation-test.el
```diff
@@ -68,6 +68,7 @@ values of customisable variables."
            (search-forward "|")
            (delete-char -1)
            (clojure-mode)
+           (font-lock-ensure)
```
bbatsov: How can this affect an indentation test?

carlosgeos: I think it is because the indentation command behaves differently depending on whether font lock mode is ON or OFF, in the case of docstrings.

This branch is master with the test added: https://github.com/carlosgeos/clojure-mode/tree/temp-branch. The tests are fine, but if `(font-lock-ensure)` is added, the test with the docstring fails (like it should).

bbatsov: Hmm, that’s odd. Turning on the mode should enable the font locking.

carlosgeos: Usually yes, but apparently not when using temp-buffer:

https://stackoverflow.com/a/23569337
https://emacs.stackexchange.com/a/44301

carlosgeos:
True, that example is incorrectly font locked. I'll make some modifications.

carlosgeos:
The new solution handles whitespace and newlines better

If everything is ok, I will squash

clojure-mode.el
```diff
@@ -955,7 +955,13 @@ highlighted region)."
                              (setq docelt (1- docelt)))))
                        (and (zerop docelt) (<= (point) startpos)
                             (progn (forward-comment (point-max)) t)
-                            (= (point) (nth 8 state)))))
+                            (= (point) (nth 8 state))))
+                     ;; In a def, at last position is not a docstring
+                     (not (and (string= "def" firstsym)
+                               (save-excursion
+                                 (goto-char startpos)
```
cichli: Is this necessary given the `goto-char` below?

cichli:
Apart from that one final comment, everything looks great now.

clojure-mode.el
carlosgeos: The first `goto-char` looks redundant but I think it is not, because there is a `(sexp-at-point)` below, and it needs the point at the correct position beforehand.

Also, the `(sexp-at-point)` form does not take any arguments so I cannot pass `startpos` to it.

Anyway, I tried without it and it didn't work. Happy to remove it if we find the way to make it work.

cichli: Ah, great point. I didn't think of that! Yeah, we definitely need it because of the `(sexp-at-point)`.

cichli: Go ahead and squash if you like. I will merge once I have the rights to do so.

carlosgeos:
Thanks, I also think you can close https://github.com/clojure-emacs/clojure-mode/issues/469 and https://github.com/clojure-emacs/clojure-mode/issues/405

PR #507, opened 2019-01-21
[Fix #506] fix clojure-mode-display-version

### LINKED ISSUE #506: clojure-mode-display-version not working
## Expected behavior

Running `clojure-mode-display-version` should return:

## Actual behavior

Returns

`clojure-mode (version nil)`

Start Emacs. Open a .clj buffer. Run `clojure-mode-display-version`

wish I could

### Emacs version

26.1

### Operating system

Biebian

carlosgeos:
`lm-version` was returning nil, since it tried to look at the buffer headers to get the version, but it couldn't during load. I gave the contents of `clojure-mode.el`, in a temporary buffer to (lm-version) so it can read it.

Perhaps all the contents of `clojure-mode.el` should not be copied over to the temporary buffer. Should this buffer contain only part of it (only the file headers for example) ?

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Bugs fixed
 
 * Dynamic vars whose names contain non-alphanumeric characters are now font-locked correctly.
+* [#506](https://github.com/clojure-emacs/clojure-mode/issues/506): `clojure-mode-display-version` correctly displays the package's version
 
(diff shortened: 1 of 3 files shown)


carlosgeos:
Actually, `Version` is fine. This doesn't fix anything.

My `(lm-version)` returns `nil` on the first call always.

PR #509, opened 2019-01-21
[Fix #508] correct font-lock for namespace name

### LINKED ISSUE #508: [font-lock] namespace name after shorthand metadata not working
## Expected behavior
```clojure
(ns ^:md myns) ;; <-- myns coloured in font-lock-type-face
```
## Actual behavior
```clojure
(ns ^:md myns) ;; <-- not font-lock-type-face
```

Just type it

### Emacs version

GNU Emacs 26.1 (build 1, x86_64-apple-darwin18.2.0, Carbon Version 158 AppKit 1671.2)
 of 2019-01-03

carlosgeos:
Things like:

```clojure
(ns ^:some_md ^{:more "yes"} ^:ok my.ns)
```
now give the correct font-lock to `my.ns`

I just tweaked the regex to accept metadata in shorthand notation

-----------------


clojure-mode.el
```diff
@@ -861,8 +861,8 @@ any number of matches of `clojure--sym-forbidden-rest-chars'."))
 
       ;; namespace definitions: (ns foo.bar)
       (,(concat "(\\<ns\\>[ \r\n\t]*"
-                ;; Possibly metadata
-                "\\(?:\\^?{[^}]+}[ \r\n\t]*\\)*"
+                ;; Possibly metadata, shorthand and/or longhand
+                "\\(?:\\^?\\(?:{[^}]+}\\|:[^ \r\n\t]+[ \r\n\t]\\)[ \r\n\t]*\\)*"
```
cichli: Could we use `\s-` or `[:space:]` instead of `[ \r\n\t]`? Does it matter? It seems we use all three in different places throughout the code.

carlosgeos: Certainly, I just saw `[ \r\n\t]` being used and chose that. I'll change it and see if it causes any problems

carlosgeos:
Yep, will squash

carlosgeos:
@cichli I kept `[ \r\n\t]` because I had some problems testing with `[:space:]` and `\s-`. Not sure why though, since `[:space:]` should include the newline characters, but I did.

cichli:
Great work! Thanks again. 🙌

PR #510, opened 2019-01-21
[Fix comment in #445] Proper font lock in (s/def ::keyword)

carlosgeos:
Fixes what is said here: https://github.com/clojure-emacs/clojure-mode/issues/445#issuecomment-340460753. The regex could be improved so feedback appreciated. Basically, it has to exclude words starting with `:` so it does not apply `function-name-face` to them (they are keywords)

-----------------


cichli:
Thanks again for this set of PRs, it's greatly appreciated! 🙌

PR #514, opened 2019-02-22
Initial attempt to run on CircleCI

danielcompton:
Converts the CI config to use CircleCI. You can see an example of this running at https://circleci.com/gh/danielcompton/clojure-mode/tree/use-circle-ci.

A few notes:

* CircleCI has a feature called orbs which can reduce duplication further, but I don't know enough about the commonalities to abstract things further at the moment.
* I can add more Emacs versions to test against, but 25 latest, 26 latest, and master seemed like decent coverage. I don't have enough knowledge here really.
* I tried running `make test-checks` in CI, but it didn't work. I'm not sure how/if that should be brought over? 
* CircleCI doesn't have the concept of allowable failures, I've included emacs master, but that will probably break on occasions.

-----------------

Thanks!


bbatsov:
Nice! You also have to install cask in the default steps, otherwise you can’t run the tests or the test checks.

danielcompton:
Cask is already installed in those Docker images when you use the -dev variants. I think the tests are running correctly, although the output looks a little different than on Travis.

bbatsov:
Oh, I didn’t know this. So, I guess we only really need to update the readme icon.

.circleci/config.yml
```diff
@@ -0,0 +1,39 @@
...
+      - image: silex/emacs:26-dev
+        entrypoint: bash
+    <<: *default-steps
+
+  test-emacs-master:
+    docker:
+      - image: silex/emacs:master-dev
+        entrypoint: bash
+    <<: *default-steps
+#
```
cichli: One tiny nit: there should be another newline above this for consistency, and we can probably elide the `#` if we don't actually need a comment.

.circleci/config.yml
```diff
@@ -0,0 +1,39 @@
+version: 2.0
```
cichli: 2.1 is also available now.

cichli:
@bbatsov could you enable `Build forked pull requests` [here](https://circleci.com/gh/clojure-emacs/clojure-mode/edit#advanced-settings)?

danielcompton:
Thanks for the review, I think I've addressed all of the issues.

>> I tried running `make test-checks` in CI, but it didn't work. I'm not sure how/if that should be brought over?

> Could you link to the logs?

https://circleci.com/gh/danielcompton/clojure-mode/12

cichli:
Thanks! Looks like `test-checks` [checks any .el file](https://github.com/clojure-emacs/clojure-mode/blob/master/test/test-checks.el#L18-L19) including the `-autoloads.el` and `-pkg.el` files if they're present. I think we could exclude them in that regex. On Travis we always ran this task in a fresh environment so it never saw those two files.

bbatsov:
@cichli Now you’ve got the power.

danielcompton:
I've changed the ordering of test-checks to come before test-bytecomp. I tried to adjust the regular expression to exclude the generated files, but Emacs doesn't support lookahead regular expressions, so I don't think it's possible, at least not without writing another function to filter the list afterwards.

cichli:
Thanks for this! 🙌

PR #515, opened 2019-02-28
[Fix #506] Makes display version command return the actual version (2)

### LINKED ISSUE #506: clojure-mode-display-version not working

carlosgeos:
My previous PR (https://github.com/clojure-emacs/clojure-mode/pull/507), only partially solves the problem (https://github.com/clojure-emacs/clojure-mode/issues/506)

The byte-compiled file will look for the version in its headers but it has none.

The version has to be given to the byte compiled file at compile time (therefore using `eval-when-compile`).

**Difference in the .elc files:**
This PR:
screenshot_20190228_022353

master branch:
screenshot_20190228_022436

---
And a small change: `(lm-version)` takes an argument, the file. This is cleaner than creating a temp buffer, inserting the contents of the current file and calling (lm-version) within this context.

-----------------

### DIFF
--- a/clojure-mode.el
+++ b/clojure-mode.el
@@ -82,9 +82,8 @@
   :link '(emacs-commentary-link :tag "Commentary" "clojure-mode"))
 
 (defconst clojure-mode-version
-  (let ((thisbuffer (or load-file-name buffer-file-name)))
-    (with-temp-buffer (insert-file-contents thisbuffer)
-                      (lm-version)))
+  (eval-when-compile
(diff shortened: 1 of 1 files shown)


bbatsov:
The commit message doesn’t seem right. I guess you can use a variation of your PR description instead.

PR #519, opened 2019-03-23
Fix clojure-find-ns when there's a ns form inside a string

Malabarba:

-----------------

Thanks!

### DIFF
--- a/clojure-mode.el
+++ b/clojure-mode.el
@@ -1886,6 +1886,19 @@ the cached value will be updated automatically."
 (defvar-local clojure-cached-ns nil
   "A buffer ns cache used to speed up ns-related operations.")
 
+(defun clojure--find-ns-in-direction (direction)
+  "Return the nearest namespace in a specific DIRECTION.
+DIRECTION is `forward' or `backward'."
+  (let ((candidate)
+        (fn (if (eq direction 'forward)
+                #'search-forward-regexp
+              #'search-backward-regexp)))
+    (while (and (not candidate)
+                (funcall fn clojure-namespace-name-regex nil t))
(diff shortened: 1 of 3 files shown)


clojure-mode.el
```diff
@@ -1894,19 +1907,16 @@ no namespaces above point, return the first one in the buffer.
...
-                  ;; Move to top-level to avoid searching from inside ns
-                  (ignore-errors (while t (up-list nil t t)))
-
-                  ;; The closest ns form above point.
-                  (when (or (re-search-backward clojure-namespace-name-regex nil t)
-                            ;; Or any form at all.
-                            (and (goto-char (point-min))
-                                 (re-search-forward clojure-namespace-name-regex nil t)))
-                    (match-string-no-properties 4))))))
+    (let* ((ns (save-excursion
```
bbatsov: Seems to me a regular `let` would do here.

clojure-mode.el
```diff
@@ -1886,6 +1886,19 @@ the cached value will be updated automatically."
 (defvar-local clojure-cached-ns nil
   "A buffer ns cache used to speed up ns-related operations.")
 
+(defun clojure--find-ns-in-direction (dir)
```
bbatsov: Every time I see `dir` I think of directory, but that's not a big deal. It's a good idea to extract this.

clojure-mode.el
Malabarba: done

clojure-mode.el
Malabarba: No reason to be lazy anyway. Changed it to direction.

bbatsov:
Thanks!

Issue #494, opened 2018-10-14
Fix compilation warnings


bbatsov:
```
Compiling /home/jonas/.emacs.d/lib/clojure-mode/clojure-mode.el...

In toplevel form:
clojure-mode.el:476:1:Warning: Unused lexical argument ‘args’
clojure-mode.el:476:1:Warning: Unused lexical argument ‘orig-fun’

In clojure--replace-let-bindings-and-indent:
clojure-mode.el:487:12:Warning: looking-back called with 1 argument, but
    requires 2-3
clojure-mode.el:2023:1:Warning: Unused lexical variable
    ‘clojure-comment-start’
clojure-mode.el:2083:1:Warning: Unused lexical variable ‘beg’

In clojure-cycle-not:
clojure-mode.el:2426:10:Warning: looking-back called with 1 argument, but
    requires 2-3
clojure-mode.el:2640:1:Warning: Unused lexical variable ‘k’
```

See https://github.com/tarsius/hl-todo/issues/30 for details.

PR #521, opened 2019-04-09
Fix compilation warnings  #494

rpkarlsson:
Fix compilation warnings as stated in #494.

-----------------

### DIFF
--- a/clojure-mode.el
+++ b/clojure-mode.el
@@ -475,18 +475,11 @@ ENDP and DELIMITER."
 (declare-function paredit-close-curly "ext:paredit")
 (declare-function paredit-convolute-sexp "ext:paredit")
 
-(defun clojure--replace-let-bindings-and-indent (orig-fun &rest args)
-  "Advise ORIG-FUN to replace let bindings.
-
-Sexps are replace by their bound name if a let form was
-convoluted.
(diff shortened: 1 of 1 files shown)


clojure-mode.el
```diff
@@ -2063,9 +2056,8 @@ many times."
         (condition-case nil
             (save-match-data
               (let ((original-position (point))
-                    clojure-comment-start clojure-comment-end)
+                    clojure-comment-end)
                 (beginning-of-defun)
-                (setq clojure-comment-start (point))
```
bbatsov: Why did you change this?

rpkarlsson: `clojure-comment-start` was generating a warning and as far as I can tell It is unused? It may be that I'm missunderstanding something?
```
clojure-mode.el:2023:1:Warning: Unused lexical variable
    ‘clojure-comment-start’
```

bbatsov: Hmm, I think that's supposed to be a buffer local variable. Let me take a closer look at the code - we might have made some real mistake. :-)

bbatsov:
Sorry about the long radio silence! After a closer inspection of the changes I feel they are all safe. Thanks!

PR #522, opened 2019-04-09
[Fix #496] Highlight [[var]] style comments

### LINKED ISSUE #496: Improve highlighting of wikilinks and markdown code elements
`[[wikilinks]]` are commonly used to refer to other vars or namespaces in docstrings. [cljdoc supports them](https://github.com/cljdoc/cljdoc/blob/master/doc/userguide/for-library-authors.adoc#wikilinks) (Codox does too) and highlighting those more could be nice. Perhaps the same formatting as regular backtick-code would be just fine.

I also noticed that strings in backticks are only "highlighted" if there's no space in the backtick-wrapped string. Perhaps this is something worth keeping in mind when tackling this issue.

rpkarlsson:
Highlight [[var]] style comments as suggested in  #496 using the same rules as backtick-code highlighting.
 -----------------


bbatsov:
Looks good to me, but you'll need to add also a couple of unit tests and a changelog entry.

bbatsov:
Thanks!

PR #534, opened 2019-07-10
[Fix #531] Don't match strings when using `clojure-rename-ns-alias`.

### LINKED ISSUE #531: Scoping issues with clojure-rename-ns-alias
The new `clojure-rename-ns-alias` refactoring introduced in #529 should only modify namespaced symbols/keywords, but the current regexp is too broad and could unintentionally match with things like filenames and slashes used in prose to indicate choices.
 
I'm not sure what the expected behaviour should be, perhaps a good idea would be to skip matching of strings but not comments, since strings found in source code don't generally refer to interned symbols (apart from references in docstrings where the full unaliased namespaces would be used)

Another option might be to delegate matching strings and comments to `query-replace` so that such cases can be spotted and dealt with interactively.

```clojure
(ns my.ns
  (:require [my.lib :as lib]))

(def dirname "/usr/local/lib/python3.6/site-packages")

;; TODO: refactor using lib/foo
(def data (lib/baz dirname))
```

- Execute `M-x clojure-rename-ns-alias lib RET new-lib RET`

Result (note the invalid directory name):

```clojure
(ns my.ns
  (:require [my.lib :as new-lib]))

(def dirname "/usr/local/new-lib/python3.6/site-packages")

;; TODO refactor using new-lib/foo
(def data (new-lib/baz dirname))
```

```
clojure-mode (version 5.11.0-snapshot)
```

### Emacs version
26.2

### Operating system
macOS 10.14.5

anthonygalea:
Is it acceptable to use font locking to determine whether a string was matched?

### DIFF
--- a/clojure-mode.el
+++ b/clojure-mode.el
@@ -2662,13 +2662,12 @@ lists up."
   (clojure--find-ns-in-direction 'backward)
   (let ((rgx (concat ":as +" current-alias))
         (bound (save-excursion (forward-list 1) (point))))
-    (when (save-excursion (search-forward-regexp rgx bound t))
-      (save-excursion
-        (while (re-search-forward rgx bound t)
-          (replace-match (concat ":as " new-alias))))
+    (when (search-forward-regexp rgx bound t)
+      (replace-match (concat ":as " new-alias))
       (save-excursion
         (while (re-search-forward (concat current-alias "/") nil t)
-          (replace-match (concat new-alias "/"))))
+          (when (not (nth 3 (syntax-ppss)))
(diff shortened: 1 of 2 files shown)


bbatsov:
> Is it acceptable to use font locking to determine whether a string was matched?

Generally it's better to use `syntax-ppss` (https://www.gnu.org/software/emacs/manual/html_node/elisp/Position-Parse.html)

anthonygalea:
thanks @bbatsov

bbatsov:
Thank you!

PR #538, opened 2019-07-12
Add default keybindings for rename-ns-alias and add-arity refactorings, and refactor clojure-unwind

yuhan0:
I placed these suggested bindings under the clojure-refactor-map and added them to the menubar:
`C-c C-r n r` -> `clojure-rename-ns-alias`
`C-c C-r a` -> `clojure-add-arity`

The `a` keybinding was taken by `clojure-unwind-all` which didn't made much sense, so I took the liberty of changing it to `U` as a "stronger" version of the existing `u -> clojure-unwind` binding.

-----------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,8 +5,8 @@
 ### New features
 
 * [#496](https://github.com/clojure-emacs/clojure-mode/issues/496): Highlight `[[wikilinks]]` in comments.
-* [#366](https://github.com/clojure-emacs/clj-refactor.el/issues/366): Add support for renaming ns aliases (`clojure-rename-ns-alias`).
-* [#410](https://github.com/clojure-emacs/clojure-mode/issues/410): Add support for adding an arity to a function (`clojure-add-arity`).
+* [#366](https://github.com/clojure-emacs/clj-refactor.el/issues/366): Add support for renaming ns aliases (`clojure-rename-ns-alias`, default binding `C-c C-r n r`).
+* [#410](https://github.com/clojure-emacs/clojure-mode/issues/410): Add support for adding an arity to a function (`clojure-add-arity`, default binding `C-c C-r a`), .
 
 ### Bugs fixed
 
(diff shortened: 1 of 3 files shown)


bbatsov:
> The a keybinding was taken by clojure-unwind-all which didn't made much sense, so I took the liberty of changing it to U as a "stronger" version of the existing u -> clojure-unwind binding.

Uppercase bindings are both hard to press and a bit confusing for the end users, so I generally avoid them. Probably in this case we can fold a few of the `*-all` commands into the parent commands and make the invokable with a prefix arg.

yuhan0:
Would you like that to be done in this PR? It sounds like quite a big breaking change and personally I prefer having separate commands for discoverability instead of using prefix args. 

Another option in this case could be `u w -> unwind` and `u a -> unwind-all`, or `u` for one and `C-u` for the other, although that would somewhat break symmetry with the rest of the bindings.

bbatsov:
> Would you like that to be done in this PR? It sounds like quite a big breaking change and personally I prefer having separate commands for discoverability instead of using prefix args.

That makes sense under a certain point - when you start running out of meaningful keybindings you know you have to start compacting things and group them together. I'm not saying to remove the existing command - I'm just saying to remove its dedicated keybinding. I think such commands are a classic use case for a prefix key as technically this is just an extension of the core behaviour. 

Changing the existing key is just as big of a breaking change as far as I'm concerned, so nothing really changes in the end. 

> Another option in this case could be u w -> unwind and u a -> unwind-all, or u for one and C-u for the other, although that would somewhat break symmetry with the rest of the bindings.

Extra prefixes should be utilized only to group multiple related commands. E.g. - if we had more commands that operated on function definitions it would make sense to put them under some shared prefix.

yuhan0:
Ok, I refactored `clojure-unwind` to take a numeric prefix / universal argument, and removed the dedicated key binding for `clojure-unwind-all`

bbatsov:
The changes look great! Don't forget to update the changelog as well.

yuhan0:
Done, and added a couple of tests as well :)

bbatsov:
Lovely! Thank you! 🙇

PR #540, opened 2019-07-20
[FIX #520] allow clojure-align-cond-forms to recognize qualified forms

### LINKED ISSUE #520: Allow clojure-align-cond-forms to recognize namespace-qualified forms
This is useful for adding user defined functions to the list of `clojure-align-cond-forms` and then calling them from within another ns.

## Expected behavior
Proper indentation:
```
(clojure.core/case :foo
  :foo       "foo"
  :bar       "bar"
  :foobarbaz "foobarbaz")
```
## Actual behavior
Actual indentation is: 
```
(clojure.core/case :foo
  :foo "foo"
  :bar "bar"
  :foobarbaz "foobarbaz")
```
Take the above code block and run `clojure-align`. Notice that it does not align as non-ns-qualified `case` would.

```
clojure-mode (version 5.8.2)
```

### Emacs version

`25.3.1`

### Operating system

`macOS 10.13.4`

wandersoncferreira:
I only extended the list of values at `clojure-align-cond-forms` to include the **qualified** versions of each symbol. Should be the most straightforward way to solve this issue.

-----------------

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### Bugs fixed
+
+* [#520](https://github.com/clojure-emacs/clojure-mode/issues/508): Fix allow `clojure-align-cond-forms` to recognize qualified forms.
+
 ## 5.11.0 (2019-07-16)
 
 ### New features
--- a/clojure-mode.el
+++ b/clojure-mode.el
@@ -1130,7 +1130,9 @@ will align the values like this:
   :safe #'listp
   :type '(repeat string))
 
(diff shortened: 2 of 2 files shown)


bbatsov:
The solution looks reasonable to me, although the repetition annoys me a bit. @Malabarba is this fine by you?

bbatsov:
Guess there are no objections from him.

Malabarba:
Sorry for the delay. I'm still managing my inbox overflow.
Yes, the change looks fine. Anything smarter than this would be beyond the scope of clojure-mode anyway.

PR #541, opened 2019-07-23
Enhance the add arity refactoring

anthonygalea:
Just came across a case which the current `clojure-add-arity` refactoring does not handle well i.e. when a `defn` lies within a reader conditional. This pull request modifies `clojure-add-arity` to support this case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@
 ### Bugs fixed
 
 * [#520](https://github.com/clojure-emacs/clojure-mode/issues/508): Fix allow `clojure-align-cond-forms` to recognize qualified forms.
+* Enhance add arity refactoring to support a defn inside a reader conditional.
+
+### Changes
+
+* Enhance add arity refactoring to support new forms: letfn, fn, defmacro, defmethod, defprotocol, reify and proxy.
 
 ## 5.11.0 (2019-07-16)
 
--- a/clojure-mode.el
+++ b/clojure-mode.el
@@ -2738,37 +2738,80 @@ With a numeric prefix argument the let is introduced N lists up."
               (clojure--rename-ns-alias-internal current-alias new-alias))
           (message "Cannot find namespace alias: '%s'" current-alias))))))
 
-;;;###autoload
-(defun clojure-add-arity ()
-  "Add an arity to a function."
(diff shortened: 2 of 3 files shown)


bbatsov:
We'll have to mention the bugfix in the changelog as well. Btw, does the command handle properly `letfn` and things like `proxy`, `reify`, `defprotocol`?

anthonygalea:
Can you think of any other cases @bbatsov?

yuhan0:
What about `defmacro`? And to some extent`defmethod`, although adding an arity to a single method probably isn't very useful. 
Also the anonymous `fn`form can take multiple arities and would be nice if it was supported too :)

anthonygalea:
Many thanks for your feedback @yuhan0. Wasn't even aware that it was possible to have multi-arity in a `defmethod`.

clojure-mode.el
```diff
@@ -2738,37 +2738,80 @@ With a numeric prefix argument the let is introduced N lists up."
...
 
-;;;###autoload
-(defun clojure-add-arity ()
-  "Add an arity to a function."
-  (interactive)
-  (let ((end (save-excursion (end-of-defun)
-                             (point)))
-        (beg (progn (beginning-of-defun)
-                    (point))))
+(defun clojure-add-arity-defprotocol-internal ()
```
bbatsov: Private functions are typically named `clojure--something`.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Bugs fixed
 
 * [#520](https://github.com/clojure-emacs/clojure-mode/issues/508): Fix allow `clojure-align-cond-forms` to recognize qualified forms.
+* Enhance add arity refactoring to support reader conditionals, letfn, fn, defmacro, defmethod, defprotocol, reify and proxy in addition to defn.
```
bbatsov: I'd split this to an entry about the bugfix and a different one about adding support for more forms.

bbatsov:
Nicely done! Now I've started wonder if we need a remove arity counterpart. :-)

anthonygalea:
Hadn't thought of that. Guessing it will add very little value, but we'll never know until it's implemented. I'll give it a shot.

Issue #404, opened 2016-08-21
Syntax highlighting for #_ forms comments out next form


danielcompton:
## Expected behavior

When using the #_ reader macro, you can chain several together. This lets you comment out more than one form at a time. However the Cider syntax highlighting uses the first #_ to comment out the second one. For example:

In Cursive:

<img width="172" alt="screenshot of intellij idea 12-08-16 8-08-51 am" src="https://cloud.githubusercontent.com/assets/811954/17603155/036b2d9e-6064-11e6-8ba7-2adff68ee7bc.png">

## Actual behavior

In emacs:

<img width="220" alt="screenshot of emacs 12-08-16 8-05-56 am" src="https://cloud.githubusercontent.com/assets/811954/17603138/ece6f40e-6063-11e6-84d3-7fd03e3e6058.png">

## Steps to reproduce the problem

```clj
#_ #_ :a :b
```

### clojure-mode version information

*Include here the version string displayed by `M-x
clojure-mode-display-version`. Here's an example:*

```
clojure-mode (version 5.5.2)
```

### Emacs version

`25.1.50.1`

### Operating system

OS X 10.11.6

Carried over from https://github.com/clojure-emacs/cider/issues/1822

MalloZup:
autogenerated with https://github.com/MalloZup/doghub:  issue inactive since 450 days. Please update the issue or close it

yuhan0:
This should be closed by #545 too :)

PR #545, opened 2019-10-01
Support multiple consecutive reader comments (#_#_a b)

sulami:
Modified the default regexps and the heuristic to find the end of the region to
comment out.

Previously Emacs would treat the second `#_` as the commented form, and then
highlight the following two forms as usual. Now it (mostly) matches what Clojure
actually evaluates. Things get weird when you start mixing `#_` and forms, but
this fixes the most common use cases, like key-value-pairs.

Tagging #404/#528 for visibility

**Note:** I couldn't get the tests to work locally, so I'd be grateful if someone wants to add a test case for the above.

-----------------

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 * [#520](https://github.com/clojure-emacs/clojure-mode/issues/508): Fix allow `clojure-align-cond-forms` to recognize qualified forms.
 * Enhance add arity refactoring to support a defn inside a reader conditional.
+* [#404](https://github.com/clojure-emacs/clojure-mode/issues/404)/[#528]((https://github.com/clojure-emacs/clojure-mode/issues/528)) Fix syntax highlighting for multiple consecutive comment reader macros (`#_#_`)
 
 ### Changes
 
--- a/clojure-mode.el
+++ b/clojure-mode.el
@@ -654,14 +654,16 @@ If JUSTIFY is non-nil, justify as well as fill the paragraph."
 ;; Code heavily borrowed from Slime.
 ;; https://github.com/slime/slime/blob/master/contrib/slime-fontifying-fu.el#L186
 (defvar clojure--comment-macro-regexp
-  (rx "#_" (* " ") (group-n 1 (not (any " "))))
+  (rx (seq (+ (seq "#_" (* " ")))) (group-n 1 (not (any " "))))
   "Regexp matching the start of a comment sexp.
 The beginning of match-group 1 should be before the sexp to be
 marked as a comment.  The end of sexp is found with
 `clojure-forward-logical-sexp'.")
 
 (defvar clojure--reader-and-comment-regexp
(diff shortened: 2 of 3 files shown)


bbatsov:
> Note: I couldn't get the tests to work locally, so I'd be grateful if someone wants to add a test case for the above.

What's the problem you had with the tests? Make test works fine for me.

sulami:
> > Note: I couldn't get the tests to work locally, so I'd be grateful if someone wants to add a test case for the above.

> What's the problem you had with the tests? Make test works fine for me.

```
make: *** No rule to make target `/Users/sulami/.emacs.d/straight/repos/clojure-mode/.cask/26.3/elpa', needed by `test'.  Stop.
```

bbatsov:
What happens when you do `emacs --version` in a terminal? 

Perhaps you can run `make elpa` before `make test` and this will work?

sulami:
Emacs 26.3 (+ licensing info)

`make elpa && make test` worked fine (and passed), thanks! I'll have a look if I can craft a test for my change.

sulami:
Okay, I've added some tests. It font-locks the second pair in  `#_#_` to comment if there's nothing following, which isn't strictly true in the "what the clojure reader sees" sense, but it's no like `#_` by itself is valid anything anyway.

yuhan0:
Hi, just as a point of comparison, I had my own implementation of this feature roughly as follows, which uses a "stack" in the search function to determine the discarded range instead of regex.

```elisp
(defun clojure--search-comment-macro-internal (limit)
  "Search for a comment forward stopping at LIMIT."
  (when (search-forward-regexp clojure-comment-regexp limit t)
    (let* ((md (match-data))
           (start (match-beginning 1))
           (state (syntax-ppss start))
           (skip 1))
      ;; inside string or comment?
      (if (or (nth 3 state)
              (nth 4 state))
          (clojure--search-comment-macro-internal limit)
        (goto-char start)
        (save-match-data
          (while (< 0 skip)
            (if (looking-at clojure-comment-regexp)
                (progn (goto-char (match-beginning 1))
                       (incf skip))
              (clojure-forward-logical-sexp 1)
              (skip-chars-forward " \n")
              (decf skip))))
        ;; Data for (match-end 1).
        (setf (elt md 2) (elt md 0))
        (setf (elt md 3) (point))
        (set-match-data md)
        t))))
```

The (dubious) advantage of this is that it works when you have interleaving forms , better emulating how the clojure reader works.
image

However It might be much slower than this implementation which is good for 99.9% of use cases (not sure how font-lock performance could be benchmarked)

bbatsov:
> (not sure how font-lock performance could be benchmarked)

I guess you can just generate a huge Clojure buffer, start the Emacs profiler and see how much time will be spend in font-locking with both approaches.

bbatsov:
@sulami Thanks for tackling this! Sorry about the slow turnaround.

PR #546, opened 2019-10-30
Font lock additions

Macroz:
I would like to propose the following changes to font locking. They make it really easy to emphasize important (top level) definitions by e.g. font-size or other. I have been using this for a couple years already and polished it up a bit to be hopefully accepted. Thank you for your consideration.

possibilities

- Add specific customizable faces of different definition types.
- Support `definterface` like `defprotocol` (seems to be missing).
- Support Schema `defschema` as a type definition.

The customizable faces are
- `clojure-function-definition-face`,
- `clojure-variable-definition-face`,
- `clojure-type-definition-face`,
- `clojure-keyword-definition-face` and
- `clojure-namespace-definition-face`.

The new faces have the same default face as before that they inherit.

I wonder if this should also be included in the README but it is readily available in the wonderful Emacs customize feature.

-----------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### New features
+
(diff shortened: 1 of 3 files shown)


Macroz:
Looks like there is probably not enough interest for this and there are some obvious problems too so I'll close this PR. If anyone is interested you can ping me, I'm using this.

Issue #360, opened 2016-01-15
Align a nested map


rfkm:
Run `clojure-align` against the following code:

```clj
{:a {:a :a
     :bbbb :b}
 :bbbb :b}
```

Then you'll get:

```clj
{:a    {:a :a
     :bbbb :b}
 :bbbb :b}
```

Is this intended behavior?
`align-cljlet` can do this in one step:
```clj
{:a    {:a    :a
        :bbbb :b}
 :bbbb :b}
```

bbatsov:
Yeah, there's room for improvement here. @Malabarba will you, please, look into this?

Malabarba:
Of course!

rfkm:
Sorry, I lied.
`align-cljlet` can not do this in _one_ step.
Actually, it requires two steps:

```clj
{:a {:a :a
     :bbbb :b}
 :bbbb :b}

{:a    {:a :a
        :bbbb :b}
 :bbbb :b}

{:a    {:a    :a
        :bbbb :b}
 :bbbb :b}
```

This behavior is acceptable to me. (Of course, it would be nice if the recursive version is _optionally_ available.)
So, the only problem is `clojure-align` doesn't indent after alignment?

Malabarba:
> So, the only problem is clojure-align doesn't indent after alignment?

Yes, exactly.

rfkm:
@Malabarba 
Great, thanks! I've confirmed the fix works.

However, I've found another problem...

Run `clojure-align` against the following map:

```clj
{:a {:a :a
     :aaaaaaaaaaaaaaaaaaaaaa :a}
 :b {:a :a
     :aa :a}}
```

Then:

```clj
{:a {:a                      :a
     :aaaaaaaaaaaaaaaaaaaaaa :a}
 :b {:a :a  ; <-
     :aa :a}}
```

It works with the following map:

```clj
{:a {:a :a
     :aaaaaaaaaaaa :a}  ; <- decrease the number of `a`
 :b {:a :a
     :aa :a}}
```

Probably, `end` passed to `clojure--find-sexp-to-align` should be a marker?

bbatsov:
> Run clojure-align against the following map:

Btw, it's way more convenient to use `TAB` to align. In my mind this functionality was always intended to be used like this (smarter indentation versus the needed to realign manually with the explicit command).

rfkm:
[...]

PR #552, opened 2020-01-24
[Fix #551] Indent `clojure-align` region before aligning.

### LINKED ISSUE #551: `clojure-align` misaligns code that is improperly indented
# Expected behavior

Using `clojure-align` on a block like this, which is not properly indented to
begin with:

```
(let [a-long-name 10
b 20])
```

Should result in properly indented and aligned code, like this:

```
(let [a-long-name 10
      b           20])
```

## Actual behavior

A single `clojure-align` application results in this:

```
(let [a-long-name 10
      b                 20])
```

A second application arrives at the proper result.

This is due to the `clojure-align` implementation, which first aligns the region as-is, then performs an indentation step.

Use `clojure-align` on improperly indented blocks of code, as above.

`clojure-mode-20191112.1948`

### Emacs version

Emacs 26.3

### Operating system

Linux

kylc:
This prevents issues when aligning improperly indented code. Before, the region
would be aligned while retaining the improper indentation. The follow-up
indentation step would then break the alignment that was just performed.

With this commit, we switch the order by indenting first then aligning. This
retains compatibility with the test cases in #360.

-----------------

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Bugs fixed
 
+* [#551](https://github.com/clojure-emacs/clojure-mode/issues/551): Indent `clojure-align` region before aligning.
 * [#520](https://github.com/clojure-emacs/clojure-mode/issues/508): Fix allow `clojure-align-cond-forms` to recognize qualified forms.
 * Enhance add arity refactoring to support a defn inside a reader conditional.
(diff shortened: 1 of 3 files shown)


bbatsov:
The change looks good, but it might be a good idea to add an unit test as well.

kylc:
I've updated the commit to include a unit test for the case described in the bug report.

Thanks for taking the time to review!

bbatsov:
Sorry for the slow turnaround here. Didn't have much time for clojure-mode recently.

PR #555, opened 2020-03-20
Fix ns detection for complex metadata

yuhan0:
See https://github.com/clojure-emacs/cider/issues/2821

Use clojure-forward-logical-sexp to reliably skip metadata instead of error-prone regex.

The previous variable `clojure-namespace-name-regex` is kept as obsolete in case it breaks other packages - can remove it directly if that's preferred.

One of the commented out `TODO`  tests used a old metadata syntax `#^{:bar true}` which wasn't supported by `clojure--looking-at-non-logical-sexp` 

-----------------

### DIFF
--- a/clojure-mode.el
+++ b/clojure-mode.el
@@ -1888,6 +1888,11 @@ content) are considered part of the preceding sexp."
       (zero-or-one (any ":'")) ;; (in-ns 'foo) or (ns+ :user)
       (group (one-or-more (not (any "()\"" whitespace))) symbol-end)))
 
+(make-obsolete-variable 'clojure-namespace-name-regex 'clojure-namespace-regexp "5.12.0")
+
+(defconst clojure-namespace-regexp
+  (rx line-start "(" (? "clojure.core/") (or "in-ns" "ns" "ns+")))
+
 (defcustom clojure-cache-ns nil
   "Whether to cache the results of `clojure-find-ns'.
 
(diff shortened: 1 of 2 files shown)


clojure-mode.el
```diff
@@ -1888,6 +1888,11 @@ content) are considered part of the preceding sexp."
       (zero-or-one (any ":'")) ;; (in-ns 'foo) or (ns+ :user)
       (group (one-or-more (not (any "()\"" whitespace))) symbol-end)))
 
+(make-obsolete-variable 'clojure-namespace-name-regex 'clojure-namespace-regexp "5.12.0")
+
+(defconst clojure-namespace-regexp
```
bbatsov: I think you've added this accidentally.

yuhan0: I wasn't sure whether to change the current variable's meaning or introduce a new variable while keeping the old one around and marking it as obsolete in case some other packages currently depend on it. What do you think?

bbatsov: Ah, I didn't see you actually use this. Well, I guess it's fine like this. When I scanned quickly the code I thought it had the same name as the old one. :-)

bbatsov:
Nicely done!

yuhan0:
Thanks! Just as another data point from the wild, the `clojure.string` ns form was being detected wrongly before due to the docstring containing "(ns ...":
https://github.com/clojure/clojure/blob/master/src/clj/clojure/string.clj#L14

With the changes in this PR it now works correctly :)

PR #556, opened 2020-03-20
Rename ns alias improvements

yuhan0:
Some minor improvements to the command `clojure-rename-ns-alias`, offering minibuffer completion of existing aliases and using regexp-quote, so that namespaces containing characters like `*` or `+` can be renamed correctly without need for manual escaping.

-----------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,14 @@
 
 ## master (unreleased)
 
+### New features
+
+* [#556](https://github.com/clojure-emacs/clojure-mode/issues/556): `clojure-rename-ns-alias` picks up existing aliases for minibuffer completion.
+
 ### Bugs fixed
 
+* [#556](https://github.com/clojure-emacs/clojure-mode/issues/556): Fix renaming of ns aliases containing regex characters.
+* [#555](https://github.com/clojure-emacs/clojure-mode/issues/555): Fix ns detection for ns forms with complex metadata.
 * [#550](https://github.com/clojure-emacs/clojure-mode/issues/550): Fix `outline-regexp` so `outline-insert-heading` behaves correctly.
 * [#551](https://github.com/clojure-emacs/clojure-mode/issues/551): Indent `clojure-align` region before aligning.
 * [#520](https://github.com/clojure-emacs/clojure-mode/issues/508): Fix allow `clojure-align-cond-forms` to recognize qualified forms.
--- a/clojure-mode.el
+++ b/clojure-mode.el
@@ -2686,16 +2686,16 @@ lists up."
 (defun clojure--rename-ns-alias-internal (current-alias new-alias)
   "Rename a namespace alias CURRENT-ALIAS to NEW-ALIAS."
   (clojure--find-ns-in-direction 'backward)
-  (let ((rgx (concat ":as +" current-alias))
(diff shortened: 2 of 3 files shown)


clojure-mode.el
```diff
@@ -2684,6 +2684,19 @@ lists up."
     (insert sexp)
     (clojure--replace-sexps-with-bindings-and-indent)))
 
+(defun clojure--collect-ns-aliases ()
```
bbatsov: I guess this one can even be public API and optionally take a `ns` form as a param.

clojure-mode.el
```diff
@@ -2684,6 +2684,19 @@ lists up."
...
+  (save-excursion
+    (clojure--find-ns-in-direction 'backward)
+    (let ((end (cdr (bounds-of-thing-at-point 'list)))
+          (rgx (rx ":as" (+ space)
+                   (group-n 1 (+ (not (in " ,]\n"))))))
+          (res ()))
+      (while (re-search-forward rgx end 'noerror)
+        (message (match-string 1))
+        (push (cons (point) (match-string-no-properties 1)) res))
+      res)))
```
bbatsov: I'm not sure I understand the value of returning cons cells here if the first item in the cons cell is not the name the alias is referring to.

CHANGELOG.md
```diff
@@ -2,8 +2,12 @@
 
 ## master (unreleased)
 
-### Bugs fixed
+### New features
+* [#556](https://github.com/clojure-emacs/clojure-mode/issues/556): `clojure-rename-ns-alias` picks up existing aliases for minibuffer completion.
```
bbatsov: Please, add a blank line after each heading.

clojure-mode.el
yuhan0: Do you mean having it as a string->list<string> pure function? Or taking a `point` param marking the location of the start of the ns form?

clojure-mode.el
yuhan0: Hmm, I wrote this a few months ago and can't remember why I had that in the first place, will simplify

clojure-mode.el
bbatsov: Pure function taking a ns form as param.

clojure-mode.el
bbatsov: 👍

yuhan0:
I discovered and fixed another bug when aliases have common prefixes, eg.
```clj
[clojure.string :as string]
[clojure.spec.alpha :as s]
```

renaming s -> x would produce `[clojure.string :as xtring]`

Refactored `clojure-collect-ns-aliases` into a pure fn as requested, although it does seem a little un-idiomatic for elisp.

I can squash related commits together if needed, thanks

bbatsov:
> Refactored clojure-collect-ns-aliases into a pure fn as requested, although it does seem a little un-idiomatic for elisp.

Looks fine to me. Add a couple of unit tests for it and we're in business. 

> I discovered and fixed another bug when aliases have common prefixes, eg.

Great!

> I can squash related commits together if needed, thanks

That'd be ideal! Thanks!

yuhan0:
Okay, squashed the commits and included one test for the collect-aliases function.

bbatsov:
Thanks!

PR #557, opened 2020-03-21
Fix clojure--sym-regexp search for namespace symbols

dmitry-t7ko:
This PR demonstrates a fix for `cider-repl-set-ns` command. 

Issue appeared with namespaces in the form of e.g. `aoc-2019.puzzles.day14`.
For a file within this namespace inner function `clojure--find-ns-in-direction` would return `.puzzles.day14`, which would render cider repl unusable.

Emacs version: GNU Emacs 26.3 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.10) of 2019-08-29
OS: linux
clojure-mode version: `M-x clojure-mode-display-version` reports `clojure-mode (version nil)`. 
It was installed as part of doom-emacs, HEAD is at 
`64ed043 * origin/master update find-ns tests`

### DIFF
--- a/clojure-mode.el
+++ b/clojure-mode.el
@@ -1922,7 +1922,8 @@ DIRECTION is `forward' or `backward'."
           (save-match-data
             (goto-char end)
             (clojure-forward-logical-sexp)
-            (when (and (looking-back clojure--sym-regexp end 'greedy)
+            (clojure-backward-logical-sexp)
+            (when (and (looking-at clojure--sym-regexp)
                        (not (clojure--in-string-p))
                        (not (clojure--in-comment-p)))
               (setq candidate (match-string-no-properties 0)))))))


dmitry-t7ko:
Not complete fix, apparently - some test fail. I'll try looking at that tomorrow, but I hope I outlined the original problem clearly enough.

yuhan0:
Whoops, sorry for that regression!  Your fix looks about right, just use backward-sexp to go to beginning of symbol instead of clojure-backward-logical-sexp.

yuhan0:
Or directly using thing-at-point:
```clj
(defun clojure--find-ns-in-direction (direction)
  "Return the nearest namespace in a specific DIRECTION.
DIRECTION is `forward' or `backward'."
  (let ((candidate)
        (fn (if (eq direction 'forward)
                #'search-forward-regexp
              #'search-backward-regexp)))
    (while (and (not candidate)
                (funcall fn clojure-namespace-regexp nil t))
      (let ((end (match-end 0)))
        (save-excursion
          (save-match-data
            (goto-char end)
            (clojure-forward-logical-sexp)
            (unless (or (clojure--in-string-p) (clojure--in-comment-p))
              (setq candidate (thing-at-point 'symbol)))))))
    candidate))
```

bbatsov:
I'm fine with both approaches. Might be best to leave the patch to @yuhan0, as he's more familiar with the code.

yuhan0:
Thanks, I opened a separate PR

PR #558, opened 2020-03-21
Fix ns detection with namespaces containing containing numbers

yuhan0:
Use thing-at-point instead of `looking-back`. See #557

-----------------

Thanks!


bbatsov:
Thanks!

Issue #270, opened 2015-01-11
Decouple from lisp-mode


bbatsov:
Clojure is reasonably different from both Emacs Lisp and Common Lisp and we should decouple clojure-mode completely from `lisp-mode.el`. The fact that a lot of code is being leveraged from there makes implementing certain things pretty hard.

andreas-roehler:
IMO stuff like outline-regexp should not be hard-coded by clojure-mode-variables, but made customizable.
Emacs want be able to tweak the mode - so let's make that easy. 

Interested?

bbatsov:
While this is not a defcustom everyone can customize it even now (using the `clojure-mode-hook`). After all those are just buffer-local-variables.

andreas-roehler:
Using a hook requires some specific Emacs Lisp knowledge, while M-x customize... is understood by beginners.

bbatsov:
> Using a hook requires some specific Emacs Lisp knowledge, while M-x customize... is understood by beginners.

That's really debatable. :-)

MalloZup:
autogenerated with https://github.com/MalloZup/doghub:  issue inactive since 450 days. Please update the issue or close it

PR #559, opened 2020-03-26
Inline and fix syntax table

yuhan0:
In Clojure single quotes `'` can be part of a symbol literal anywhere except the head, this is stated explicitly in https://clojure.org/reference/reader and recently confirmed by Alex Miller on the Clojurians Slack.

Changing the syntax of `'` in the syntax table affects the output of functions like `(thing-at-point 'symbol)` which currently does not treat quotes as part of the symbol. This in turn affects `cider-symbol-at-point` and dependent functions like`cider-doc`, which currently does not work on vars named eg. `foo'` or `hawai'i`

I also took the liberty of inlining the definition of clojure-mode's syntax table rather than inheriting and overriding bits from `emacs-lisp-mode` (which in turn inherits from `lisp-mode`), making it simpler to understand it as a whole.

Also added commas as whitespace to the syntax table, this helps with some edge cases where `[:space:]` is used in a font-lock regex.

-----------------

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 
 ### Changes
 
+* Inline definition of `clojure-mode-syntax-table` and support `'` quotes in symbols and commas as whitespace.
 * Enhance add arity refactoring to support a `defn` inside a reader conditional.
 * Enhance add arity refactoring to support new forms: `letfn`, `fn`, `defmacro`, `defmethod`, `defprotocol`, `reify` and `proxy`.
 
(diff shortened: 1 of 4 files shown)


bbatsov:
Great changes!

> I also took the liberty of inlining the definition of clojure-mode's syntax table rather than inheriting and overriding bits from emacs-lisp-mode (which in turn inherits from lisp-mode), making it simpler to understand it as a whole.

I've been meaning to this for a very long time, so it's the right move. (see #270) 

You'll have to look into the build failures and add some changelog entry.

yuhan0:
Fixed the emacs-25 build failure, I didn't realise string-trim's optional argument was a new feature in 26. The emacs-master failure was there in previous PRs and seems to be some unrelated checkdoc bug. 

I also just realised that this would introduce a breakage in the current `cider-symbol-at-point` when called on quoted symbols. I could open a PR for Cider with my changes, hopefully users who upgrade packages would do both together?

bbatsov:
I guess they do, so this shouldn't be a big problem.

bbatsov:
Thanks for tackling this!

PR #566, opened 2020-04-18
Fix auto-insertion of spaces in paredit

### LINKED ISSUE #565: paredit-mode inserts extra space in clojure-mode
I'm not sure whether the following is a clojure-mode or paredit-mode bug. Recent updates seem to have introduced this, however, and it doesn't happen when using paredit-mode in other modes (e.g., emacs-lisp-mode).

## Expected behavior

If I want to quote a list, I would typically type `'(` and paredit-mode would insert the closing parenthesis and it would show `'()` in the buffer.

## Actual behavior

In clojure-mode, when I type the opening parenthesis, paredit-mode inserts an extra space between the tick and the opening parenthesis, leaving `'<space>()` rather than `'()` in the buffer (where `<space>` represents the extra space). This does NOT happen when using paredit-mode with emacs-lisp-mode, lisp-mode, or scheme-mode. This suggests that the root cause of the bug is located in clojure-mode.

The following init.el shows a minimal configuration. With this init.el, start emacs and then in the `*scratch*` buffer, type `'(` in emacs-lisp mode. Notice that no extra space is entered between the tick and the opening parenthesis.

Next, switch to clojure-mode (e.g., just use `M-x clojure-mode` in the `*scratch*` buffer). Type `'(` again and notice the extra space.

Sample init.el:

```
(require 'package)
(add-to-list 'package-archives
             '("melpa" . "http://melpa.milkbox.net/packages/")
             t)
(package-initialize)

;;; Bootstrap use-package if it isn't already installed
(unless (package-installed-p 'use-package)
  (unless package-archive-contents
    (package-refresh-contents))
  (package-install 'use-package))

(require 'use-package)

(use-package paredit
  :ensure t
  :hook ((emacs-lisp-mode clojure-mode cider-repl-mode lisp-mode scheme-mode) . paredit-mode))

(use-package clojure-mode
  :ensure t
  :mode (("\\.clj\\'" . clojure-mode)
         ("\\.cljs\\'" . clojurescript-mode)
         ("\\.cljc\\'" . clojurec-mode)
         ("\\.cljx\\'" . clojurex-mode)))
```

### clojure-mode version

```
clojure-mode (version 5.2.0-snapshot)
```

### Emacs version

```
GNU Emacs 26.3 (build 1, x86_64-apple-darwin18.2.0, NS appkit-1671.20 Version 10.14.3 (Build 18D109))
 of 2019-09-02
```

installed via `brew cask install emacs`.

### Operating system

MacOS 10.15.4 (Catalina)

yuhan0:
Fixes #565 

The first 3 commits should be squashed together, I just found the original function hard to understand and had to refactor it along the way and then remove what seemed like uneccessary checks.

-----------------

### DIFF
--- a/clojure-mode.el
+++ b/clojure-mode.el
@@ -452,14 +452,17 @@ The command will prompt you to select one of the available sections."
   (interactive)
   (browse-url clojure-style-guide-url))
 
-(defun clojure-space-for-delimiter-p (endp _delim)
+(defun clojure-space-for-delimiter-p (endp delim)
   "Prevent paredit from inserting useless spaces.
 See `paredit-space-for-delimiter-predicates' for the meaning of
 ENDP and DELIM."
   (and (not endp)
-       ;; don't insert after ' or # that is part of auto-gensym or reader conditional syntax
-       (not (looking-back "\\_<\\(?:'+\\|#\\??\\)" (point-at-bol)))))
-
+       ;; don't insert after opening quotes, auto-gensym syntax, or reader tags
+       (not (looking-back
+             (if (member delim clojure-omit-space-between-tag-and-delimiters)
+                 "\\_<\\(?:'+\\|#.*\\)"
+               "\\_<\\(?:'+\\|#\\)")
+             (point-at-bol)))))
 
 (defconst clojure--collection-tag-regexp "#\\(::[a-zA-Z0-9._-]*\\|:?\\([a-zA-Z0-9._-]+/\\)?[a-zA-Z0-9._-]+\\)"
   "Collection reader macro tag regexp.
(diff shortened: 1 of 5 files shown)


clojure-mode.el
```diff
@@ -457,20 +457,11 @@ The command will prompt you to select one of the available sections."
 See `paredit-space-for-delimiter-predicates' for the meaning of
 ENDP and DELIM."
   (or endp
-      (not (memq delim '(?\" ?{ ?\( )))
-      (not (or (derived-mode-p 'clojure-mode)
+      (not (memq delim '(?\" ?\{ ?\()))  ;;  always insert for [
+      (not (or (derived-mode-p 'clojure-mode) ;; FIXME why does this check exist
```
bbatsov: Because without it this code would be applied to any mode where `paredit` is enabled.

yuhan0: Ah, I thought it was a mode local variable - will add it back in.

yuhan0: Actually, wouldn't it make more sense to do `(make-local-variable 'paredit-space-for-delimiter-predicates)` inside clojure-paredit-setup rather than add global predicates?

test/clojure-mode-external-interaction-test.el
```diff
@@ -0,0 +1,108 @@
...
+;;; Code:
+
+(require 'clojure-mode)
+(require 'buttercup)
+(require 'paredit)
+
+(describe "Interactions with Paredit"
+  ;; reuse existing when-refactoring-it macro
+  (describe "it should insert a space"
+    (when-refactoring-it "before lists"
```
bbatsov, line 29: I get the point of using this macro here, but probably we need some alias for it, otherwise the tests read a bit weird. Probably not a big deal, though, as I can't suggest a great name right now - something like `with-edits/actions`.

clojure-mode.el
bbatsov: I guess that's an option as well.

yuhan0:
I did some more refactoring, realising that the other function `clojure-no-space-after-tag` was mostly repeated logic, and could be folded into a single predicate.

`clojure--collection-tag-regexp` doesn't seem to be used elsewhere so I made it obsolete too.
Side note: it also had the bug of using a naive `"[a-zA-Z0-9]"` regex which does not adequately describe clojure symbols. I noticed a few other regexes in the file which have a similar issue, but  that's probably best left for another PR

yuhan0:
Also I'm not a Paredit user myself (which is why this issue originally went unnoticed), so it would be great if someone could help test these changes out!

test/clojure-mode-external-interaction-test.el
yuhan0: I couldn't think of a name as well, but at least the console output makes sense -

```
Interactions with Paredit:
  it should insert a space
    before lists (0.19ms)
    before vectors (0.17ms)
    before maps (0.16ms)
    before strings (0.15ms)
    after gensym (0.20ms)
    after symbols ending with ' (0.19ms)
  it should not insert a space
    for anonymous fn syntax (0.19ms)
    for hash sets (0.17ms)
    for regexes (0.16ms)
    for quoted collections (0.17ms)
    for reader conditionals (0.17ms)

```

bbatsov:
> I noticed a few other regexes in the file which have a similar issue, but that's probably best left for another PR

Agreed.

> Also I'm not a Paredit user myself (which is why this issue originally went unnoticed), so it would be great if someone could help test these changes out!

I guess you're into smartparens or lispy, right? Given the unit tests I'm reasonably confident with your changes, so I'll just merge them and see what feedback we'll get. I'm using paredit, but I guess I don't update my Emacs packages often enough if I didn't notice this myself. :D

PR #573, opened 2020-09-28
Reference :defn in docstring, not defun

nivekuil:
This was changed in
https://github.com/clojure-emacs/clojure-mode/commit/99629deca7938e5a3f359ea3c3896c4a6ca3f5ef
but the docstring wasn't updated


bbatsov:
Thanks!

PR #579, opened 2020-11-26
Remove calls to `redisplay` from `clojure-sort-ns`

tomdl89:
This avoids unwanted flicker when jumping from current location
to ns form and back. Especially useful when `clojure-sort-ns` is
called from `before-save-hook`, and would otherwise distract the user.

-----------------

Thanks!

### DIFF
--- a/clojure-mode.el
+++ b/clojure-mode.el
@@ -1827,7 +1827,6 @@ content) are considered part of the preceding sexp."
   (if (clojure-find-ns)
       (save-excursion
         (goto-char (match-beginning 0))
-        (redisplay)
         (let ((beg (point))
               (ns))
           (forward-sexp 1)
@@ -1842,10 +1841,7 @@ content) are considered part of the preceding sexp."
           (goto-char beg)
           (if (looking-at (regexp-quote ns))
               (message "ns form is already sorted")
-            (sleep-for 0.1)
-            (redisplay)
-            (message "ns form has been sorted")
-            (sleep-for 0.1))))
+            (message "ns form has been sorted"))))
     (user-error "Can't find ns form")))
 
 (defconst clojure-namespace-name-regex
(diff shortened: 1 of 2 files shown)


bbatsov:
Won't some `save-excursion` also tackle this? 

> Especially useful when clojure-sort-ns is
called from before-save-hook, and would otherwise distract the user.

Probably you should document this somewhere, otherwise you'll probably be the only user of the new functionality. :-)

tomdl89:
The `(redisplay)` that jumps us up to the ns occurs _within_ a `save-excursion` (see new ln 1829) so I'm not sure I understand how adding more `save-excursion`s would fix this. Good point on the documentation. I should add that to the docstring.

bbatsov:
> The (redisplay) that jumps us up to the ns occurs within a save-excursion (see new ln 1829) so I'm not sure I understand how adding more save-excursions would fix this.

I wrote this without really reading the code. :-) I'm not quite sure what's the nature of the flicker you experience. From the `redisplay`'s docstring it seems that the cursor shouldn't jump around, but I'm not familiar with the function.

tomdl89:
The cursor jumps around because of the `goto-char` command (and then again because of the `forward-sexp` etc commands). Usually, save-excursion allows all of this to be done without redisplaying (so the user never sees the movement), however, I guess whoever wrote this code thought that the user would be reassured seeing the cursor jump up to the ns form and then back down again. I guess it confirms that something is happening? I added the `suppress-redisplay?` arg to allow the user to opt out. I would be fine with removing the `redisplay` calls, but presumably some users like to see the jumps? (When I wrote "flicker" I was just referring to fact that the jumps are so fast, btw).

bbatsov:
I'd just remove the `redisplay` if it's not adding anything meaningful. I always prefer to delete code rather than add more code to deal with the existing code. :-)

tomdl89:
Ok :) that was my preference all along, so suits me! Thanks

clojure-mode.el
```diff
@@ -1843,7 +1842,6 @@ content) are considered part of the preceding sexp."
           (if (looking-at (regexp-quote ns))
               (message "ns form is already sorted")
             (sleep-for 0.1)
-            (redisplay)
```
bbatsov, line 1846: I think you shoudl also kill those surrounding sleeps, as I can't see why they would be needed.

bbatsov: I see one before and after the redisplay.

bbatsov:
Thanks!

PR #582, opened 2021-02-19
feat: make indentation of special arguments customisable

hugoduncan:
Allow control of the indentation of special arguments through a defcustom.

- introduces clojure-special-arg-indent-factor, which is used as a factor of
lisp-body-ident to indent special arguments.  Defaults to 2, the currently hard
coded value for this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -220,6 +220,20 @@ The indentation of special arguments is controlled by
 `clojure-special-arg-indent-factor`, which by default indents special arguments
 a further `lisp-body-indent` when compared to ordinary arguments.
 
+An example of the default formatting is:
+
+```clojure
+(defrecord MyRecord
+    [my-field])
+```
+
+Setting `clojure-special-arg-indent-factor` to 1, results in:
+
+```clojure
+(defrecord MyRecord
+  [my-field])
+```
+
 ### Indentation of Comments
 
 `clojure-mode` differentiates between comments like `;`, `;;`, etc.
--- a/README.md
+++ b/README.md
@@ -216,6 +216,10 @@ For instructions on how to write these specifications, see
 [this document](https://docs.cider.mx/cider/indent_spec.html).
 The only difference is that you're allowed to use lists instead of vectors.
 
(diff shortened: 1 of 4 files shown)


bbatsov:
What problem is this aiming to solve? I don't recall ever seeing someone use a different indent level for macro forms.

hugoduncan:
At work we recently started using zprint to format code.  zprint doesn't use a different indent for special arguments.  This PR would allow achieving identical formatting with clojure-mode.

bbatsov:
Looking at the examples it seems to the that such forms use 2-space indent there as well https://github.com/kkinnear/zprint/blob/master/doc/options/community.md Can you share some examples so that I can understand better the difference? 

I guess you should add some examples to the README as well, if we want others to learn about this option and its potential usefulness.

hugoduncan:
I don't see a relevant example in the zprint link above.  The formatting in question is the handling of special arguments when they are on separate lines.  I've added an example with `defrecord` to the README.

bbatsov:
Ah, now I see what you mean. I always place the special args on the first line, so it took me a while to process why something like this needed to be customized. I'll ruminate a bit more on this and I'll likely merge it over the weekend.

hugoduncan:
Let me know if there is anything that I can do to facilitate this.

README.md
```diff
@@ -216,6 +216,24 @@ For instructions on how to write these specifications, see
 [this document](https://docs.cider.mx/cider/indent_spec.html).
 The only difference is that you're allowed to use lists instead of vectors.
 
+The indentation of special arguments is controlled by
```
bbatsov, line 219: I think this explanation to be a bit confusing, because I don't think we mention the meaning of special arguments anywhere in the readme and at least for me - the special argument is this case (and most of the time) is the body. I understand the examples, but I'm wondering if others won't be confused by them.

bbatsov: Ah, forget about this - https://docs.cider.mx/cider/indent_spec.html#special-arguments I no longer remember some of the things I wrote myself. :D

Issue #411, opened 2016-10-16
Add #_ reader form


benedekfazekas:
Migrated over from clojure-emacs/clj-refactor.el#334 originally created by @agzam
Agreement was that this should rather come here. see more discussion on the original issue.

the original description is as follows:
Clojure's `#_` (Ignore next form reader macro) is very useful. Would be very cool to have a function and a dedicated key binding to toggle it.

My initial attempt looks like this:

```
(defun cljr-toggle-ignore-next-form ()
  "Ignore next form (#_)"
  (interactive)
  (if (search-backward "#_" 2 t 1)
  (delete-char 2)
(progn
  (let ((fc (following-char)))
    (cond ((-contains? '( ?\) ?\] ?\} ) fc) (paredit-backward-up))
      ((-contains? '( ?\( ?\[ ?\: ?\{ ) fc) nil)
      (t (beginning-of-thing 'sexp)))
    (insert "#_")))))
```

vspinu:
Two notes to consider if someone ever implements this:

 - When deleting `#_` don't search beyond current top level form.
 - Would be nice to have cycling behavior. That is, if cljr-toggle-ignore-next-form is pressed multiple times in a row, the `#_` mark gets pushed one enclosing sexp up. 

Without the second item, I don't see much use of this, it's straitforward to get one level up and type #_ anyways.

yuhan0:
Hi, I wrote these functions a while ago to toggle the ignore form for the surrounding list or symbol below the cursor: 


```elisp
(defun clojure--toggle-ignore-next-sexp ()
  "Insert or delete the `#_' characters at the current point.
Assumes point is directly before a sexp or the #_ characters"
  (if (or (looking-at "#_")
          (search-backward "#_" (- (point) 2) t 1)
          (and (looking-at-p "_") (search-backward "#" (- (point) 2) t 1)))
      (delete-char 2)
    (insert "#_")))

(defun clojure-toggle-ignore-form (&optional arg)
  "Toggle the #_ ignore reader form for the surrounding form at point.
With optional ARG, move up by ARG surrounding lists first.
When called with C-u, act on the current top level defun."
  (interactive "P")
  (save-excursion
    (if (consp arg) ;; called with C-u
        (beginning-of-defun)
      (unless (looking-at-p "[({[]")
        ;; guard against moving past top level
        (condition-case nil
            (backward-up-list arg t t)
          (scan-error nil))))
[...]

PR #583, opened 2021-02-22
Toggle ignore reader forms

### LINKED ISSUE #567: Feature: command(s) for inserting/deleting #_ ignore forms
It would be useful to have some convenient way of inserting/deleting #_ forms built into clojure-mode, but its implementation may depend on personal preference. I hope this issue attracts some discussion as to what would suit the majority of users.

My personal implementation of this idea can be found here, which I've been tweaking over the past year: https://gist.github.com/yuhan0/38ac43c764017917fc3f3e5dc9104f1b
There are two interactive commands defined there,  `clojure-toggle-ignore-form` and `clojure-toggle-ignore-surrounding-form`- see docstrings for their usage.

I made some usability choices in the way it interprets numeric and universal arguments, and there's some degree of DWIM by wrapping an active region in `(comment)` forms and inserting blank lines before top-level forms to reduce diffs.

But it's clear that this may not suit everyone, in particular see 
https://github.com/dotemacs/clojure-comment-dwim.el/issues which takes a more DWIM approach with a single entry point and more logic to determine what the command should do in different contexts and when repeatedly called.

See also https://github.com/BetterThanTomorrow/calva/issues/505
for another interpretation of what such a command may do in different scenarios (note it's more selection-based as VSCode doesn't have numeric command arguments)

yuhan0:
Closes #567, #411

Adds 3 new interactive commands: 
`clojure-toggle-ignore`, `clojure-toggle-ignore-surrounding-form`, and `clojure-toggle-defun`

I added a bunch of tests which should demonstrate how the commands work.
No default keybindings added yet, I'm not sure if this should count as a "refactoring" and go under the `C-c C-r` prefix? 

Note about naming: the clojure.org guide calls it "discard" (https://clojure.org/guides/weird_characters), but I've seen it mostly referred to as the "ignore" form elsewhere.

-----------------

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### New features
+
+* [#567](https://github.com/clojure-emacs/clojure-mode/issues/567): Add new commands `clojure-toggle-ignore`, `clojure-toggle-ignore-surrounding-form`, and `clojure-toggle-defun` for inserting/deleting #_ ignore forms.
+
 ### Changes
 
 * [#571](https://github.com/clojure-emacs/clojure-mode/issues/571): Remove `project.el` integration.
--- a/clojure-mode.el
+++ b/clojure-mode.el
@@ -246,6 +246,10 @@ Out-of-the box `clojure-mode' understands lein, boot, gradle,
     (define-key map (kbd "s b") #'clojure-let-backward-slurp-sexp)
     (define-key map (kbd "C-a") #'clojure-add-arity)
     (define-key map (kbd "a") #'clojure-add-arity)
+    (define-key map (kbd "-") #'clojure-toggle-ignore)
+    (define-key map (kbd "C--") #'clojure-toggle-ignore)
+    (define-key map (kbd "_") #'clojure-toggle-ignore-surrounding-form)
+    (define-key map (kbd "C-_") #'clojure-toggle-ignore-surrounding-form)
     map)
   "Keymap for Clojure refactoring commands.")
 (fset 'clojure-refactor-map clojure-refactor-map)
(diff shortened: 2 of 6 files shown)


bbatsov:
> No default keybindings added yet, I'm not sure if this should count as a "refactoring" and go under the C-c C-r prefix?

I don't think that's a refactoring, so no need to use the `C-c C-r` keymap.

yuhan0:
Do you have suggestions for good default keybindings? I personally use evil-mode with the mappings 
`<leader>it` -> clojure-toggle-ignore
`<leader>is` -> clojure-toggle-ignore-surrounding-form (C-u for ignore-defun)

Just realised that the family of "refactorings" `clojure-cycle-*`, `clojure-convert-collection-*`  have a similar scope to this, perhaps I should rename them `cycle-ignore`  instead of `toggle-ignore` and have it go under the refactor map?

`C-c C-r d` / `C-c C-r D` is available (D for discard) 
or maybe something involving `C-_` or `C--`

bbatsov:
If something has only two states - I think that "toggle" is a better name than "cycle". As for the keybinding - perhaps something with `#` or `_` in it? Haven't looked in the keymap soon, but I assume we don't have a lot of good options left.

yuhan0:
> If something has only two states - I think that "toggle" is a better name than "cycle". 

Agreed, it was more about consistency with the existing command names (`clojure-cycle-if` toggles between two states `if` and `if-not`, `clojure-cycle-privacy` between `defn` and `defn-`, etc.)

>  As for the keybinding - perhaps something with # or _ in it? Haven't looked in the keymap soon, but I assume we don't have a lot of good options left.

I find having to switching from the Ctrl to Shift key for a `#` or `_` quite awkward, but I don't use native emacs bindings regularly, so maybe I'll leave this to your judgement :)

bbatsov:
I've noticed we actually have almost no top-level keybindings so you certainly have some options. Or you can just add them to the menu and people can decide if they want to bind those. I don't have brilliant suggestions, plus I doubt people are going to use such command that much, so perhaps we're overthinking this.

yuhan0:
Ok, I added them to the menu and the refactor map under `C-c C-r -` and `C-c C-r _` respectively. 
I personally use these commands more often than any other refactoring, so it's probably just a matter of workflow style (seeing as it's been requested and reinvented several times now)

bbatsov:
Fair enough. Thanks for tackling this! We can always adjust the keybindings down the road.

PR #588, opened 2021-03-22
Fix character font lock

rvlo:
Characters escaped weren't being formatted correctly, as shown by the submitted test.

Example of code that wasn't being formatted:
["text" \\, "other"]
(str \\[ "inside" \\])

-----------------

Thanks!


rvlo:
I also propose moving the `lambda arguments` regex below `character literals`, for handling `\%` correctly. Would there be any problems in doing so?

bbatsov:
You'll have to check this. The font-locking logic is quite brittle.

The changes you proposed look reasonable. Please, mention the fix in the changelog.

rvlo:
I added a test for the % character literal, moved the `lambda arguments` below `character literals`, added an entry to the changelog, rebased on master and squashed the commits.

bbatsov:
Thanks!

PR #590, opened 2021-04-21
Rename ns aliases in selected region

yuhan0:
When copying code from other sources, the conventions for aliasing namespaces may be different (eg. `str` vs `string` for clojure.string) and one has to manually search and replace these aliases.

Since there was already a `clojure-rename-ns-alias` command with most of the relevant logic, I extended it to be able to rename aliases in a selected region. In that case, it will not touch the ns form at all and only pick up aliases from the region.

I also cleaned up the code a bit from my previous PR here, moving the ns-alias related functions together and renaming `clojure-collect-ns-aliases` with double dashes (there was really no reason for it to be public in the first place). 

The API could be changed to introduce another command instead of doing region-active DWIM on `clojure-rename-ns-alias`. But in practice I find that such renaming usually involves multiple (selected) forms, so a command that works on top-level forms probably isn't that useful.

Will update the changelog and readme after details are confirmed :) 

Thanks!

-----------------

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### New features
 
+* [#590](https://github.com/clojure-emacs/clojure-mode/pull/590): Extend `clojure-rename-ns-alias` to work on selected regions.
 * [#567](https://github.com/clojure-emacs/clojure-mode/issues/567): Add new commands `clojure-toggle-ignore`, `clojure-toggle-ignore-surrounding-form`, and `clojure-toggle-defun` for inserting/deleting #_ ignore forms.
 * [#582](https://github.com/clojure-emacs/clojure-mode/pull/582): Add `clojure-special-arg-indent-factor` to control special argument indentation.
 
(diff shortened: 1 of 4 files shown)


bbatsov:
> The API could be changed to introduce another command instead of doing region-active DWIM on clojure-rename-ns-alias. But in practice I find that such renaming usually involves multiple (selected) forms, so a command that works on top-level forms probably isn't that useful.

It's fine in its current form. 

I like the change, but it'd be nice to advertise this functionality somewhere (e.g. in the README), otherwise I doubt many people will figure out there's special handling for regions.

yuhan0:
Having used the command in some major refactorings over the past week, I also made another usability change to make it easier to rename multiple aliases at one go.
An animated GIF would be nice to illustrate the new functionality, I'm not sure if @anthonygalea would mind recording one to keep in the style of the rest of the page :)

bbatsov:
Yeah, that'd be nice indeed, but I think we're good to merge regardless. Thanks for tackling this!

anthonygalea:
Thank you for this change @yuhan0 👍 (and also for asking for the recording, happy to help)

PR #594, opened 2021-05-05
[Fix #593] clojure-find-ns breaks if the ns form is preceded by ws

### LINKED ISSUE #593: clojure-find-ns breaks if the ns form is preceded by whitespace
See https://github.com/clojure-emacs/cider/issues/3009

## Expected behavior

`clojure-find-ns` should handle properly an `ns` form with preceding whitespace.

## Actual behavior

It returns `nil`. 

```clojure
  (ns foo.bar)

(defn foo [] :bar)
```

Try evaluating the `defn` form. 

### clojure-mode version

```
clojure-mode (version 5.10.0)
```

### Emacs version

26.3

### Operating system

Ubuntu 20.04

jogo3000:
Problem was that the regexp here expected (ns form to start right in the
beginning of the line. I added zero or more whitespace in the beginning which
corrects the issue.

Adds also a unit test which covers the issue.

-----------------

- [ x] The commits are consistent with our [contribution guidelines][1].
- [ x] You've added tests (if possible) to cover your change(s). Bugfix, indentation, and font-lock tests are extremely important!
- [ x] You've run `M-x checkdoc` and fixed any warnings in the code you've written.
- [ x] You've updated the changelog (if adding/changing user-visible functionality).

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## master (unreleased)
 
+### Bugs fixed
+* [#593] clojure-find-ns breaks if the ns form is preceded by whitespace
+
 ## 5.13.0 (2021-05-05)
 
 ### New features
--- a/clojure-mode.el
+++ b/clojure-mode.el
@@ -1873,7 +1873,7 @@ content) are considered part of the preceding sexp."
 (make-obsolete-variable 'clojure-namespace-name-regex 'clojure-namespace-regexp "5.12.0")
 
 (defconst clojure-namespace-regexp
-  (rx line-start "(" (? "clojure.core/") (or "in-ns" "ns" "ns+") symbol-end))
+  (rx line-start (zero-or-more whitespace) "(" (? "clojure.core/") (or "in-ns" "ns" "ns+") symbol-end))
 
 (defcustom clojure-cache-ns nil
   "Whether to cache the results of `clojure-find-ns'.
--- a/test/clojure-mode-util-test.el
+++ b/test/clojure-mode-util-test.el
@@ -61,6 +61,8 @@
   (it "should find common namespace declarations"
     (with-clojure-buffer "(ns foo)"
       (expect (clojure-find-ns) :to-equal "foo"))
+    (with-clojure-buffer " (ns foo)"
+      (expect (clojure-find-ns) :to-equal "foo"))
     (with-clojure-buffer "(ns
     foo)"
       (expect (clojure-find-ns) :to-equal "foo"))


jogo3000:
Do you think it should handle even more wide range of cases? For example whitespace before the ns symbol? 
```
( ns foo)
```

yuhan0:
I'm not sure how to feel about this, why would there be any reason for the ns form to be preceded by whitespace or not at the top level? 
 `clojure-find-ns` is a heuristic after all, and with this change you could get false positive cases like 
```
(ns the-actual-ns) 

(comment
  (ns not-the-real-one))
```
Or even
```
(reduce (fn [[acc idx] ns]
          [(conj acc
             (ns idx))
           (inc idx)])
  [[] 0] 
  [[1 2 3] [4 5 6]])
```

yuhan0:
Also, most if not all the font-lock regexes and refactoring commands assume that an opening paren is directly followed by the function name, eg. 
https://github.com/clojure-emacs/clojure-mode/blob/33f267ac182afe8fa82cc321e9f515c0397e35f6/clojure-mode.el#L816
https://github.com/clojure-emacs/clojure-mode/blob/33f267ac182afe8fa82cc321e9f515c0397e35f6/clojure-mode.el#L1951
https://github.com/clojure-emacs/clojure-mode/blob/33f267ac182afe8fa82cc321e9f515c0397e35f6/clojure-mode.el#L2292

I think it's just a consequence of Emacs using regexes to parse language grammars, especially in lisps where these formatting conventions are pretty universally followed. Otherwise all the regexes would get quite unwieldy covering all these edge cases and even affect performance. See eg. `lisp-mode` where any opening paren at the beginning of line is treated as a top-level defun, regardless of actual structure.

bbatsov:
Technically the code is supposed to find the closest `ns` or `in-ns` form, not necessary one at the top-level. I know that using `in-ns` in source files is rare, but we have to keep it in mind.

yuhan0:
Yeah, but anything not at the top-level is statically undecidable whether it actually applies to the forms below - it could be a quoted list of forms 
```
(def core-macros
  '((if if-not or and)
    (dowhile doto dotimes)
    (ns)))
```
or something crazy like
```
(if (System/getenv "PROD-MODE")
  (in-ns 'foo.prod)
  (in-ns 'foo.dev))
```

whereas as a heuristic, a top-level form is almost guaranteed to be evaluated. 

A more rigorous way would probably involve a local var `clojure-get-ns-function` and having CIDER override it with some REPL-backed function. But I guess that's hardly ever needed outside of these contrived examples..

bbatsov:
Yep, I'm quite aware of the shortcomings of the current setup, but it has served us reasobly well for a very long time, that's why I've never been in particularly interested in improving it.

jogo3000:
Now that I see @yuhan0 's examples I see that this fix may be dangerous to merge. After all, people probably run into this issue very rarely.

yuhan0:
I tend to shadow `ns` as a plural of `n`, so the tradeoff of accidentally creating a false-positive situation verus the unlikely event of leading whitespace before an actual ns declaration doesn't seem worth it.

jogo3000:
Good catch @yuhan0 , better think about this some more.

yuhan0:
Then perhaps *only* font-locking `ns` forms at the start of the line would be more appropriate "fix", providing a visual cue that clojure-mode won't pick it up:
https://github.com/clojure-emacs/clojure-mode/blob/33f267ac182afe8fa82cc321e9f515c0397e35f6/clojure-mode.el#L874

```
"^(\\<ns\\>[ \r\n\t]*"
```

jogo3000:
I added a comment to the main issue about your findings, so it is easier to find than the conversation in this closed pull request. I hope you don't mind.

PR #596, opened 2021-06-26
Add clojure-directory-prefixes customisation

codeasone:
This change enables customisation of the auto-inserted `ns` form on new file creation. It reflects the needs of larger code projects with atypical folder structures, as for example shown below:

```
{/src,/test}
  /bb
  /clj
  /cljc
  /cljs
  /common
```
where `{:paths ["src/clj" "src/cljc" "src/cljs" "src/bb" "src/common" ...] ...}` are specified accordingly in the `deps.edn`.

With the following customisation in `dir-locals.el`, or otherwise configured:

```elisp
((nil
  . ((clojure-directory-prefixes . ("\\`clj[scx]?\\." "\\`bb\\." "\\`common\\.")))))
 ```

The correct namespace is auto-inserted for new `clj[cs]` files created within `src/bb` and `src/common`.

-----------------

- [X] The commits are consistent with our contribution guidelines.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## master (unreleased)
 
+### New features
+
 * Allow additional directories, beyond the default `clj[sc]`, to be correctly formulated by `clojure-expected-ns` via new `defcustom` entitled `clojure-directory-prefixes`
 
 ## 5.13.0 (2021-05-05)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## master (unreleased)
 
-* Allow additional eliding beyond just `clj[sc]` in `clojure-expected-ns` via new `defcustom` entitled `clojure-elided-prefixes-ns`
+* Allow additional directories, beyond the default `clj[sc]`, to be correctly formulated by `clojure-expected-ns` via new `defcustom` entitled `clojure-directory-prefixes`
 
 ## 5.13.0 (2021-05-05)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## master (unreleased)
 
+* Allow additional eliding beyond just `clj[sc]` in `clojure-expected-ns` via new `defcustom` entitled `clojure-elided-prefixes-ns`
+
 ## 5.13.0 (2021-05-05)
 
 ### New features
--- a/clojure-mode.el
+++ b/clojure-mode.el
@@ -196,9 +196,9 @@ Out-of-the box `clojure-mode' understands lein, boot, gradle,
           (and (listp value)
                (cl-every 'stringp value))))
 
-(defcustom clojure-elided-prefixes-ns
-  '()
-  "A list of additional regexp prefixes that are elided when formulating ns."
+(defcustom clojure-directory-prefixes
+  '("\\`clj[scx]?\\.")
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: clojure-mode.el [hunk 1]
```diff
@@ -196,6 +196,15 @@ Out-of-the box `clojure-mode' understands lein, boot, gradle,
           (and (listp value)
                (cl-every 'stringp value))))
 
+(defcustom clojure-elided-prefixes-ns
+  '()
+  "A list of additional regexp prefixes that are elided when formulating ns."
```
bbatsov, line 201: I think we can do better with the wording here, as I wouldn't have figured out what does this do without seeing the code change.

clojure-mode.el
```diff
@@ -1761,7 +1770,10 @@ If PATH is nil, use the path to the file backing the current buffer."
          (sans-file-sep (mapconcat 'identity (cdr (split-string sans-file-type "/")) "."))
          (sans-underscores (replace-regexp-in-string "_" "-" sans-file-sep)))
     ;; Drop prefix from ns for projects with structure src/{clj,cljs,cljc}
-    (replace-regexp-in-string "\\`clj[scx]?\\." "" sans-underscores)))
+    (cl-reduce (lambda (a x)
+                 (replace-regexp-in-string x "" a))
+               (add-to-list 'clojure-elided-prefixes-ns "\\`clj[scx]?\\.")
```
bbatsov, line 1775: Why don't you just make this the default value? (or expand it to whatever you had suggested in the message)

clojure-mode.el
bbatsov: E.g. I think it's better to think here of directory structure and just mention that `clojure-expected-ns` happens to use this.

clojure-mode.el
codeasone: Good point I'll make it the default.

clojure-mode.el
codeasone: I've renamed the option `clojure-directory-prefixes` with suggested default.

codeasone: @bbatsov I've renamed and established the default as suggested - any outstanding concerns? The patch is working well for me at the moment.

clojure-mode.el
```diff
@@ -196,6 +196,15 @@ Out-of-the box `clojure-mode' understands lein, boot, gradle,
           (and (listp value)
                (cl-every 'stringp value))))
 
+(defcustom clojure-directory-prefixes
+  '("\\`clj[scx]?\\.")
+  "A list of directory prefixes used by clojure-expected-ns to formulate correct ns."
```
bbatsov, line 201: Please, place `clojure-expected-ns` in `'`` to fix the broken CI.

clojure-mode.el
```diff
@@ -196,6 +196,15 @@ Out-of-the box `clojure-mode' understands lein, boot, gradle,
...
+  :type '(repeat string)
+  :package-version '(clojure-mode . "5.0.0")
```
bbatsov, line 203: 5.14.0

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## master (unreleased)
 
+* Allow additional directories, beyond the default `clj[sc]`, to be correctly formulated by `clojure-expected-ns` via new `defcustom` entitled `clojure-directory-prefixes`
```
bbatsov, line 5: This should go under `New features`.

PR #598, opened 2021-07-03
Mark put-clojure-indent safe for use in LocalVariables, .dir-locals.el, etc.

rlbdv:
Add a put-clojure-indent form validator and attach it as a
safe-local-eval-function property so that safe invocations won't
trigger open-file prompts when used in local variables, or when added
to .dir-locals.el like this:
```elisp
  ((clojure-mode
    (eval . (put-clojure-indent 'defrecord '(2 :form :form (1))))))
```
For now, only support specs specified as lists, not vectors.

-----------------

This work assumes that we don't want to rely on any existing validation, and so attempts to perform thorough validation up front.  Also happy to make further changes, but thought I'd post what I have, to see if I'm on a reasonable track.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,7 +26,7 @@ compile: elpa
 	$(CASK) build
 
 clean:
-	rm -f $(OBJS)
+	rm -f $(OBJS) clojure-mode-autoloads.el
 
 test: $(PKGDIR)
 	$(CASK) exec buttercup
--- a/Makefile
+++ b/Makefile
@@ -39,4 +39,4 @@ test-bytecomp: $(SRCS:.el=.elc-test)
 
 %.elc-test: %.el elpa
 	$(CASK) exec $(EMACS) --no-site-file --no-site-lisp --batch \
-		-l test/clojure-mode-bytecomp-warnings.el $
+		-l test/clojure-mode-bytecomp-warnings.el $<
--- a/clojure-mode.el
+++ b/clojure-mode.el
@@ -1845,6 +1845,9 @@ If PATH is nil, use the path to the file backing the current buffer."
   (goto-char (point-min))
   (clojure-insert-ns-form-at-point))
 
+(defvar-local clojure-cached-ns nil
+  "A buffer ns cache used to speed up ns-related operations.")
+
 (defun clojure-update-ns ()
   "Update the namespace of the current buffer.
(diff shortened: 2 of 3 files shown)


rlbdv:
I don't have access to circleci at the moment (wondering why their test results aren't public without a login), but since the `test` and `test-checks` targets work here, I wondered if it might be the `test-bytecomp` target.  That one fails on the bare `$` in the target rules, and if I do this:
```diff
--- a/Makefile
+++ b/Makefile
@@ -38,5 +38,5 @@ test-checks:
 test-bytecomp: $(SRCS:.el=.elc-test)

 %.elc-test: %.el elpa
-       $(CASK) exec $(EMACS) --no-site-file --no-site-lisp --batch \
-               -l test/clojure-mode-bytecomp-warnings.el $
+       $(CASK) exec $(EMACS) -L . --no-site-file --no-site-lisp --batch \
+               -l test/clojure-mode-bytecomp-warnings.el $<

``` 
It gets further, but fails like this:
```
In toplevel form:
clojure-mode.el:482:25:Error: reference to free variable `clojure--let-regexp'
```
After moving some functions around and removing the `#`s from `(make-obsolete #'clojure-no-space-after-tag #'clojure-space-for-delimiter-p "5.12.0")`, it gets further still, but then crashes because `end-of-thing` is undefined.  I'm just guessing that might indicate a missing `thingatpt` dependency.

rlbdv:
Not sure what's going on in CI with 25 and master, but I went ahead and fixed up everything that was causing trouble with Emacs 27 here.

rlbdv:
i.e. I moved those declarations before their first use, and added the necessary `(require 'thingatpt)`, etc.

bbatsov:
Your changes look good to me overall, but I won't be able to review them more carefully until next week. Sorry about the slow response.

rlbdv:
No worries, and no rush -- if I get time, I might see what's up with emacs 25 and master later, either locally, or in a vm.

rlbdv:
Fixed the issues.  Emacs 25 doesn't have `caddr` (switched to `nth` which is likely more approachable anyway), and test-bytecomp wouldn't allow the existing `clojure-align-separator` docstring because it's longer than 80 characters.

bbatsov:
Well done!

PR #605, opened 2021-11-14
Add `(require 'seq)`

ROCKTAKEY:
I add `(require 'seq)`, since `seq-every-p` is not autoloaded.
When it is installed by package.el, its absence causes byte-compile warning: 
```
In end of data:
clojure-mode.el:3019:1:Warning: the function ‘seq-every-p’ is not known to be
    defined.
```

-----------------

Thanks!

### DIFF
--- a/clojure-mode.el
+++ b/clojure-mode.el
@@ -1586,7 +1586,7 @@ https://docs.cider.mx/cider/indent_spec.html e.g. (2 :form
            (not (null spec))
            (or (integerp (car spec))
                (memq (car spec) '(:form :defn)))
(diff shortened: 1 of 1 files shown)


bbatsov:
Good catch. Alternatively we can just replace this with something from `cl-lib` to avoid loading the whole library for a single function.

ROCKTAKEY:
`seq-every-p` seems to be replaced with `cl-every`. Is it better?

bbatsov:
Yeah, I think that'd be a better solution for this particular problem.

ROCKTAKEY:
OK, I will fix it.

ROCKTAKEY:
Fixed.

bbatsov:
Thanks!

PR #607, opened 2022-01-03
Make it possible to get uniform indentation

### LINKED ISSUE #526: Enabling https://tonsky.me/blog/clojurefmt easily
Because of two hard-coded `always-align` calls in `clojure-indent-function` it is not as simple as `(setq clojure-indent-style 'always-indent)` to get the desired effect. As undesirable as that effect may be for some ;-)

It would be useful to have just one setting for this.

I worked around it by overriding `lisp-indent-function` this way: https://github.com/danieroux/emacs/commit/09403e601a1d0d0e33fe0d2c3238e77bef1c53b6

Changing these two `always-align` to `always-indent` also gets the desired/undesired formatting: https://github.com/clojure-emacs/clojure-mode/blob/a9cbe6a37f51da1c37035390d43500b437e8bff2/clojure-mode.el#L1499 and https://github.com/clojure-emacs/clojure-mode/blob/a9cbe6a37f51da1c37035390d43500b437e8bff2/clojure-mode.el#L1513

plexus:
While Emacs and clojure-mode have good support for "lisp-style" indentation, and
provided fine-grained customization via indent-specs, some prefer "dumb"
indentation where every single form gets the same level of indent, regardless of
whether it's a macro, whether there are arguments on the same line as the
function name, etc.

Currently this is not possible without redefining certain functions because of
hard-coded values that handle specific cases, namely the indent-spec case, and
the keyword invocation case (`(:require ...)`).

This introduces two new defcustoms, one allows completely disabling the
indent-spec handling, so the first hard-coded case is skipped, the second allows
specifically customizing the second case.

This should allow "tonsky style" formatting with

```
(setq clojure-indent-style 'always-indent
      clojure-indent-keyword-style 'always-indent
      clojure-enable-indent-specs nil)
```

Closes #526

-----------------

Thanks!

### DIFF
--- a/clojure-mode.el
+++ b/clojure-mode.el
@@ -100,6 +100,8 @@
 
 (defcustom clojure-indent-style 'always-align
   "Indentation style to use for function forms and macro forms.
+For forms that start with a keyword see `clojure-indent-keyword-style'.
+
 There are two cases of interest configured by this variable.
 
 - Case (A) is when at least one function argument is on the same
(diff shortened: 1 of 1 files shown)


plexus:
Will let this sit here for a bit while I dogfood it.

bbatsov:
I assume the people who like fixed indentation don't care about this level of granularity, so this can be just a an extra `clojure-indent-style` named `fixed/tonsky/whatever`. That's how I planned to implement this in the past, but admittedly I dropped the ball on this.

kommen:
FWIW, `clojure-ts-mode` implements it and calls it just `fixed` https://github.com/clojure-emacs/clojure-ts-mode?tab=readme-ov-file#indentation

bbatsov:
@plexus Should we merge the PR? I guess by now you had enough time to test it.

plexus:
Sure, if folks are happy with it. I don't generally like or use this style of indentation, except that one time a few years ago when a client project was using it. I don't remember much about this PR, but I guess it works?

vemv:
From memory, the only reason why @dannyfreeman made that style available in clojure-ts-mode is because it was simpler to implement, in the context of bootstrapping a functional solution.

Maybe we can use whatever influence we have to avoid a proliferation of styles. All formatters default to something traditionally emacs-y anyway. And Cursive can be trivially configured to follow the same style.

Tools have only become more handy since the infamous blog post that suggested the idea of a style that really didn't take off but still is requested from time to time. I mean, by now cljfmt can be invoked in _many_, instantaneous ways (via clojure-lsp, cider, a graalvm binary) - besides from the editor-native formatting that will typically match cljfmt.

What problem are we solving?

bbatsov:
That some people like this formatting and don't care about the alternatives? 😉 

To be fair some people liked this even before the famous blog post by Nikita - that's why `clojure-mode` supported something similar almost from day 1. (here's a bit of historical context - https://github.com/clojure-emacs/clojure-mode/commit/e3ef390fab132cac50831caab42654fef41c0da3) The style because popular not because of the blog post, but because in many teams people used different tools and this was the only style that all tools would agree on back then. But I've always considered it inferior to semantic/standard indentation and never used it myself.

Issue #608, opened 2022-01-23
Alignment: nested, multi-line sexps with margin comments result in incorrect indentation.


samwagg:
PR: https://github.com/clojure-emacs/clojure-mode/pull/609

## Expected behavior
Alignment should work correctly when nested, multi-line sexps have margin comments.
## Actual behavior
Alignment breaks.
## Steps to reproduce the problem
Try aligning the following:
```
(let [:x (+ 2 3
            3 4) ; comment 
      :y (+ 1 2)])
 ```
 Note the incorrect result:
 ```
 (let [:x         (+ 2 3
            3 4) ; comment 
      :y         (+ 1 2)])
 ```

### clojure-mode version
Hm, clojure-mode-display-version is returning `clojure-mode (version nil)`. *Packages* is showing `20201001.1449`. I have also reproduced in a unit test in the latest on master.

### Emacs version
27.2

### Operating system
mac os Big Sur 11.4

PR #609, opened 2022-01-23
Fix alignment issue involving margin comments.

samwagg:
Alignment wasn't working properly when nested, multi-line sexps were followed by margin comments.

Thanks!

### DIFF
--- a/clojure-mode.el
+++ b/clojure-mode.el
@@ -1252,6 +1252,9 @@ Place point as in `clojure--position-for-alignment'."
 
 (defun clojure--search-whitespace-after-next-sexp (&optional bound _noerror)
   "Move point after all whitespace after the next sexp.
+Additionally, move past a comment if one exists (this is only
+possible when the end of the sexp coincides with the end of a
+line).
 
 Set the match data group 1 to be this region of whitespace and
 return point.
@@ -1260,7 +1263,8 @@ BOUND is bounds the whitespace search."
   (unwind-protect
       (ignore-errors
         (clojure-forward-logical-sexp 1)
-        (search-forward-regexp "\\([,\s\t]*\\)" bound)
+        ;; Move past any whitespace or comment.
+        (search-forward-regexp "\\([,\s\t]*\\)\\(;+.*\\)?" bound)
         (pcase (syntax-after (point))
           ;; End-of-line, try again on next line.
           (`(12) (clojure--search-whitespace-after-next-sexp bound))
--- a/test/clojure-mode-indentation-test.el
+++ b/test/clojure-mode-indentation-test.el
@@ -327,43 +327,51 @@ DESCRIPTION is a string with the description of the spec."
       :a)")
 
   (when-indenting-with-point-it "should handle fixed-normal-indent"
-    "(cond
+    "
+     (cond
       (or 1
           2) 3
     |:else 4)"
 
-    "(cond
+    "
+     (cond
       (or 1
           2) 3
       |:else 4)")
 
   (when-indenting-with-point-it "should handle fixed-normal-indent-2"
-    "(fact {:spec-type
+    "
+(fact {:spec-type
        :charnock-column-id} #{\"charnock\"}
 |{:spec-type
        :charnock-column-id} #{\"current_charnock\"})"
 
-    "(fact {:spec-type
+    "
+(fact {:spec-type
        :charnock-column-id} #{\"charnock\"}
       |{:spec-type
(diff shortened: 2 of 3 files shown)


test/clojure-mode-indentation-test.el
```diff
@@ -705,6 +705,11 @@ x
     "#?@(:clj  [2]
     :cljs [2])")
 
+  (when-aligning-it "should work correctly when margin comments appear after multi-line, non-terminal sexps"
+    "(let [:x (+ 2 3
```
vemv, line 709: This seems an odd example, I really don't recall ever seeing an aligned function invocation.

Hashmaps, `case`s, etc are more frequent targets.

Also, this piece of code would look exactly the same for the aligned and not aligned cases, so one can't really tell if you're affecting alignment at all with the proposed changes.

I'd suggest choosing a variety of examples, and also, add ` ; comment`s at different positions, to get really good coverage.

samwagg: I think alignment of `let` bindings is common. I made the strange decision to use keywords for the left-hand side of the bindings, which I will definitely update.

I will update to add examples for `hashmap` and `case`. I plan to still use a single test with a single form, but I will add all three cases within the form. I think that's what you have in mind.

I will also update so that the alignment isn't trivial; I can see how that might be confusing, but I also what to make clear that it is besides the point. The point is that when you align code that is already aligned correctly, it is getting mangled ([see the attached issue](https://github.com/clojure-emacs/clojure-mode/issues/608)). This test does fail if you undo the regexp change I made.

vemv: Thanks! Appreciated.

Simply for record, I still don't see a particular alignment in the example, neither for the `let` itself or the function invocations:

image

I'm aware that `let` alignment is a thing but it's better shown e.g. here:

image

...Just in case, indentation != alignment.

btw, this initial newline I added in both of my examples would make the tests more readable. I think it would be welcome if you can update all examples to include it.

(I haven't tried running the tests like that, might require removing the initial newline programatically if present)

clojure-mode.el
```diff
@@ -1260,7 +1260,7 @@ BOUND is bounds the whitespace search."
   (unwind-protect
       (ignore-errors
         (clojure-forward-logical-sexp 1)
-        (search-forward-regexp "\\([,\s\t]*\\)" bound)
+        (search-forward-regexp "\\([,\s\t]*\\)\\(;+.*\\)?" bound)
```
vemv, line 1263: Probably at this point the regexp could be explained with a comment

samwagg:
Alright, updated. A few notes:
- I left the commits unsquashed for now in case it helps with reviewing. But I understand they need to be squashed before merge.
- I had interpreted your request to "update all examples" to include the initial newline as a request to update all the existing indentation tests. But now I'm wondering if you just meant for me to do it for my multiple examples. It seems right to be consistent, but I can easily drop that commit if desired.

vemv:
Looks all good to me! Great work.

As one last suggestion, you could add examples with `;;` comments for greater coverage.

---

I cannot further review these days for personal reasons, someone else will have to approve/merge

bbatsov:
I'll take it from here.

samwagg:
I wasn't sure if I was meant to test margin comments that begin `;;` or comments on their own line, so I added coverage for both. The latter is distinct from the issue I was testing, imo, so I added a separate test for it. I checked, and it didn't look like there were any existing tests for alignment of sexps with interleaved line comments.

samwagg:
> I wasn't sure if I was meant to test margin comments that begin `;;` or comments on their own line, so I added coverage for both. The latter is distinct from the issue I was testing, imo, so I added a separate test for it. I checked, and it didn't look like there were any existing tests for alignment of sexps with interleaved line comments.

Actually, I suppose having one test with a variety of different comment examples makes sense. I guess I just have tunnel vision for the specific bug I am fixing. I can combine the two tests into one that covers all examples if that style is preferred.

bbatsov:
More test coverage is usually a good thing, plus some conventions are not as strong with Clojure as with older Lisps - e.g. a lot of people use `;;` comments everywhere, although historically `;;` was a line comment and `;` was a margin comment.

samwagg:
Alright, I think I'm done then. I squashed my commits.

samwagg:
Hi, just dropping a reminder that I am finished making changes with this PR and I think it's ready.  Please let me know if there are any further changes desired.

bbatsov:
Thanks!

vemv:
Released as 5.14.0

PR #616, opened 2022-04-17
Support clojure-dart files

### LINKED ISSUE #615: Support clojure-dart
Support clojure-dart source files.

Currently, clojure-dart files are basically clojure source files that compiles to dart.
See:
* https://github.com/Tensegritics/ClojureDart
* https://github.com/Tensegritics/ClojureDart/blob/816344d4abaf5165fe55ebf3af662af7d3ab3ca8/doc/differences.md

## Expected behavior
Clojure-mode should support clojure-dart files (.cljd).

## Actual behavior
Current clojure-mode does not detect .cljd files.

### clojure-mode version
```
clojure-mode (version 5.14.0)
```

### Emacs version

```
28.0.91
```

### Operating system

```
MacOS 12.0.1
```

jgoday:
[Fix #615].

Adds minimal support to cljd files (clojure-dart) as a clojure derived major mode (clojuredart-mode).

-----------------

- [X] You've updated the readme (if adding/changing user-visible functionality).

Thanks!

### DIFF
--- a/clojure-mode.el
+++ b/clojure-mode.el
@@ -200,7 +200,7 @@ Out-of-the box `clojure-mode' understands lein, boot, gradle,
                (cl-every 'stringp value))))
 
 (defcustom clojure-directory-prefixes
-  '("\\`clj[scx]?\\.")
+  '("\\`clj[scxd]?\\.")
   "A list of directory prefixes used by `clojure-expected-ns'.
 The prefixes are used to generate the correct namespace."
   :type '(repeat string)
@@ -3070,7 +3070,7 @@ With universal argument \\[universal-argument], act on the \"top-level\" form."
 ;;;###autoload
 (progn
   (add-to-list 'auto-mode-alist
-               '("\\.\\(clj\\|dtm\\|edn\\)\\'" . clojure-mode))
+               '("\\.\\(clj\\|cljd\\|dtm\\|edn\\)\\'" . clojure-mode))
   (add-to-list 'auto-mode-alist '("\\.cljc\\'" . clojurec-mode))
   (add-to-list 'auto-mode-alist '("\\.cljs\\'" . clojurescript-mode))
   ;; boot build scripts are Clojure source files
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### Changes
+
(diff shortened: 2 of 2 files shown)


clojure-mode.el
```diff
@@ -3067,12 +3070,19 @@ With universal argument \\[universal-argument], act on the \"top-level\" form."
 
 \\{clojurec-mode-map}")
 
+;;;###autoload
+(define-derived-mode clojuredart-mode clojure-mode "ClojureDart"
+  "Major mode for editing ClojureDart code.
```
vemv, line 3075: Not sure if we need/want a new major mode currently, note that e.g. Babashka doesn't get one.

babashka works in clojure-mode which makes sense and has no reported issues.

Likewise `clojurec-mode` works with arbitrary reader conditionals e.g. `:cljs` or `:bb`

bbatsov: Babashka, however, is supposed to be closer to Clojure. Generally the derived modes make more sense for languages with a different runtime and syntactic oddities. I'm not familiar with ClojureDart, so it's hard to say something concrete about it, though.

ericdallo: [These are the differences](https://github.com/Tensegritics/ClojureDart/blob/main/doc/differences.md)

vemv: From what I can see in https://github.com/clojure-emacs/clojure-mode/issues/615#issuecomment-1101692937 , none of these differences so far warrants a distinct mode - they don't seem to imply changes to the things clojure-mode is concerned with e.g. syntax coloring.

It's worth noting that other target runtimes such as CLR, Erlang, etc haven't gotten a mode either.

"No is temporary, yes is forever", so we could just keep it simpler and grow it from there?

bbatsov: In practice it doesn't really matter much. The derived modes are very "cheap" and light, so it's not like they add much complexity. I added a derived mode for ClojureScript and clojurec mostly to simplify some checks in CIDER (it's easier to check the major mode than file extensions) and to get cleaner buffer grouping. As runtime like CLR and Erlang never took off there wasn't much demand to do anything about them.

jgoday: Certainly! it makes totally sense. Reverted the commit to remove clojuredart-major-mode, and just added cljd extension to auto-mode-alist clojure-mode.

bbatsov:
I see that in the meantime @jgoday simplified the PR to the bare minimum, so I guess I'll just merge this as that's too much of a discussion for a super trivial change. :-)

PR #621, opened 2022-07-06
Remove needless quote of choice values

syohex:
The customize menu of those variables does not work. Because default value `'always-align` does not match against any choice values. (The choice list is already quoted so for example  `'always-align` in the list means `''always-align`). 

Customize menu of original code is as below. It shows `(mismatch)` due to the above reason. And `[Value Menu]` does not display.

image

This PR version is below. There is a `[Value Menu]` and there is no `(mismatch)`.

image

BTW, development Emacs(29.0.50) warns this kind of mistakes by byte-compiling.

```
clojure-mode.el:101:12: Warning: defcustom for ‘clojure-indent-style’ has
    syntactically odd type ‘'(choice (const :tag Same as `lisp-mode'
    'always-align) (const :tag Indent like a macro body 'always-indent) (const
    :tag Indent like a macro body unless first arg is on the same line
    'align-arguments))’
```

-----------------

Thanks!

### DIFF
--- a/clojure-mode.el
+++ b/clojure-mode.el
@@ -141,10 +141,10 @@ to indent function forms.
           merge
           some-coll)"
   :safe #'symbolp
-  :type '(choice (const :tag "Same as `lisp-mode'" 'always-align)
-                 (const :tag "Indent like a macro body" 'always-indent)
+  :type '(choice (const :tag "Same as `lisp-mode'" always-align)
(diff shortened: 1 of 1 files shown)


vemv (approved):
LGTM!

Thanks for the screenshots, appreciated.

bbatsov:
Thanks!

Issue #622, opened 2022-07-13
some-> & some->> not fontlocked


boxxxie:
### clojure-mode version
```
clojure-mode (version 5.14.0)
```

my work around

```lisp
(defvar clojure-core-macros
  `((,(concat
       "(\\(?:clojure.core/\\)?"
       (regexp-opt
        '(
          "some->"
          "some->>"
          ) t)
       "\\>")
     1 font-lock-keyword-face)))
```

vemv:
It sounds worthwhile and cheap to support clojure.core macros though. Would be friendly to clojure-lsp users, for instance, I reckon.

wdyt?

bbatsov:
@vemv Yeah, probably those we can add back to the main package. I was wondering what to do with the extra package as I guess most people don't realize that it exists.

bbatsov:
Looking at the code we've definitely messed up something as some macros are in the main package and some in the secondary.

bbatsov:
I guess we'll need to check how many of the things that are listed as `clojure.core` built-ins in the extra package are actually macros and move them to `clojure-mode`. I'll reopen this ticket.

boxxxie:
point me to where i should look in the .el file, and i can compile a list of the missing macros from clojure.core

bbatsov:
See https://github.com/clojure-emacs/clojure-mode/blob/master/clojure-mode.el#L843

boxxxie:
macros i left out are from pprint_base.clj cl_format.clj template.clj core_proxy.clj tap.clj test.clj shell.clj junit.clj repl.clj genclass.clj main.clj

vemv:
Released as clojure-mode 5.15.0, which will be available within a couple hours

PR #623, opened 2022-07-13
ADD many missing clojure.core macros to font-lock declaration

boxxxie:
#622 

font locks added

```
"gen-class"
"while"
"import"
"pvalues"
"bound-fn"
"vswap!"
"with-loading-context"
"delay"
"gen-interface"
"with-bindings"
"deftype"
"with-precision"
"lazy-seq"
"let"
"defstruct"
"areduce"
"definline"
"future"
"fn"
"definterface"
"some->>"
"amap"
"extend-type"
"defmethod"
"time"
"memfn"
"extend-protocol"
"reify"
"defonce"
"defn-"
"defprotocol"
"sync"
"assert"
"proxy-super"
"loop"
"with-out-str"
"with-in-str"
"some->"
"proxy"
"locking"
"defmulti"
"io!"
"lazy-cat"
"comment"
"defrecord"
"refer-clojure"
```


boxxxie:
CI fail is unrelated to my changes, i dono what to do

clojure-mode.el
```diff
@@ -841,15 +841,20 @@ any number of matches of `clojure--sym-forbidden-rest-chars'."))
...
-          '("letfn" "case" "cond" "cond->" "cond->>" "condp"
-            "for" "when" "when-not" "when-let" "when-first" "when-some"
-            "if-let" "if-not" "if-some"
-            ".." "->" "->>" "as->" "doto" "and" "or"
-            "dosync" "doseq" "dotimes" "dorun" "doall"
-            "ns" "in-ns"
-            "with-open" "with-local-vars" "binding"
-            "with-redefs" "with-redefs-fn"
-            "declare") t)
+          '("->" "->>" ".." "amap" "and" "areduce" "as->" "assert" "binding"
```
vemv, line 844: While we're here we could leave use one line per item, that way it should be easier to tell whether something's there

(agreed, @bbatsov ?)

boxxxie: i prefer that @vemv, i just tried to make it look similar to how it previously looked.
i guess if @bbatsov says yes, i'll update the commit. or should i just update it and assume it'll be ok?

clojure-mode.el
```diff
@@ -841,15 +841,90 @@ any number of matches of `clojure--sym-forbidden-rest-chars'."))
...
+            "binding"
+            "bound-fn"
+            "case"
+            "comment"
+            "cond"
+            "cond->"
+            "cond->>"
+            "condp"
+            "declare"
+            "definline"
```
bbatsov, line 862: You can remove all the `def`something macros as they are handled in a special manner above. `fn` is handle there as well.

clojure-mode.el
bbatsov: I'm fine with with a name per line, although that'd makes things a bit inconsistent with the rest of the font-locking code.

bbatsov:
Don't forget to add a changelog entry about this as well.

bbatsov:
One more thing - if some of those are duplicated in `clojure-mode-extra-font-locking` they should be removed from there - e.g. `some->`.

bbatsov:
Seems good to me. Thanks!

Issue #612, opened 2022-04-01
Buffer freezes on `(def ^{:macro})`


borkdude:
## Expected behavior

When I have `(def ^{:macro})` in my buffer (as the only code), I should see no freezing of emacs.

## Actual behavior

Emacs with clojure-mode freezes.

## Steps to reproduce the problem

Put `(def ^{:macro})` as the sole code in a clojure buffer. Try to move the cursor a bit back and forth.

When Emacs freezes, send:

```
pkill -USR2 Emacs
```

and then see something like:

```
Debugger entered--entering a function:
* #f(compiled-function () #<bytecode 0x1fe4d41470e9>)()
  imenu--generic-function(((nil clojure-match-next-def 0)))
  imenu-default-create-index-function()
  imenu--make-index-alist(t)
  which-function()
  which-func-update-1(#<window 3 on foo.clj>)
  which-func-update()
  apply(which-func-update nil)
  timer-event-handler([t 0 0 500000 t which-func-update nil idle 0])
  recursive-edit()
  debug(lambda)
* thing-at-point--end-of-sexp()
  bounds-of-thing-at-point(sexp)
  clojure-match-next-def()
  imenu--generic-function(((nil clojure-match-next-def 0)))
  imenu-default-create-index-function()
  imenu--make-index-alist(t)
  which-function()
  which-func-update-1(#<window 3 on foo.clj>)
  which-func-update()
  apply(which-func-update nil)
  timer-event-handler([t 0 0 500000 t which-func-update nil idle 0])
```

@vemv hinted that this looked suspicious:

https://github.com/clojure-emacs/clojure-mode/blob/c339353f9e649b3af084f1bb6ce759e614a2f243/clojure-mode.el#L749

### clojure-mode version

```
clojure-mode (version 5.13.10)
```

### Emacs version

27.2

### Operating system

macOS

vemv:
I can repro this by invoking `(clojure-match-next-def)` in clojure-mode latest.

I could also fix it by changing here:

https://github.com/clojure-emacs/clojure-mode/blob/c339353f9e649b3af084f1bb6ce759e614a2f243/clojure-mode.el#L757-L758

```diff
-(progn (forward-sexp) (backward-sexp))
+(forward-sexp)
```

I can't make sense of `(progn (forward-sexp) (backward-sexp))`, it seems a no-op that naturally would cause an infinite loop in conjunction with the `(while (not found?)`?

[...]

PR #624, opened 2022-07-16
[Fix #595 #612] Fix infinite loop when reverse searching for next definition

### LINKED ISSUE #595: Syntax `^:` causes Emacs to freeze.
This happened to me about a half a year ago on Arch Linux with obviously older Emacs and older clojure-mode. Both using Prelude. I wasn't able to reproduce merely with clojure-mode, so steps include downloading prelude.

## Expected behavior
Adding ```^:``` to def and defn don't freeze emacs

## Actual behavior
```^:``` freezes emacs

Download Prelude Emacs. Open either a clj or cljs file. Type ```(def ^:something)``` or ```(defn ^:something)``` or even just ```(def ^:)```/```(defn ^:)```. I've reliably been able to reproduce if it doesn't happen initially by typing ```(defn ^:something)``` and then M-backspace to delete back to the colon. Will include a screen shot of cpu profiler. 

Screen Shot 2021-05-16 at 2 20 43 PM

I know a screenshot is not necessarily the best, but I'm having trouble logging it otherwise.

### clojure-mode version

clojure-mode (version 5.13.0)

### Emacs version

GNU Emacs 28.0.50 (build 1, x86_64-apple-darwin19.6.0, NS appkit-1894.60 Version 10.15.7 (Build 19H114))
 of 2021-04-24

### Operating system

MacOSX 10.15.7 Catalina

### PR DESCRIPTION
**Pull request by OknoLombarda**
> Fixes https://github.com/clojure-emacs/clojure-mode/issues/595
> Fixes https://github.com/clojure-emacs/clojure-mode/issues/612

Fix infinite loop in `clojure-match-next-def` when typing metadata in def forms. Some more context [here](https://github.com/clojure-emacs/clojure-mode/issues/612#issuecomment-1186152534).

-----------------

Thanks!

### CODE REVIEW

File: clojure-mode.el [hunk 1]
```diff
@@ -754,8 +754,10 @@ Called by `imenu--generic-function'."
               (when (char-equal ?\) (char-after (point)))
                 (backward-sexp)))
           (cl-destructuring-bind (def-beg . def-end) (bounds-of-thing-at-point 'sexp)
-            (if (char-equal ?^ (char-after def-beg))
-                (progn (forward-sexp) (backward-sexp))
+            (when (char-equal ?^ (char-after def-beg))
+              (progn (forward-sexp) (backward-sexp)))
```
vemv, line 758: While we're here, please add a comment explaining what this accomplishes

test/clojure-mode-syntax-test.el
```diff
@@ -60,6 +60,36 @@
...
+    (with-clojure-buffer "
+(def a 1)
+(list [1 2 3])"
+      (end-of-buffer)
+      (clojure-match-next-def)
+      (expect (looking-at "(def"))))
+  (it "handles empty def forms"
+    (with-clojure-buffer "
+(def)
+(list [1 2 3])"
```
vemv, line 88: Following https://github.com/clojure-emacs/clojure-mode/issues/612, there should be at least these extra cases:

* `(def ^{:macro})`
* `(def ^{:macro true})`
* `(def ^{:macro true} foo)`

And perhaps `(def ^{:macro} foo)` (which is both invalid and unlikely to type) too, especially if it doesn't complicate things.

test/clojure-mode-syntax-test.el
```diff
@@ -60,6 +60,36 @@
       (expect (non-func "^hint " form) :to-be nil)
       (expect (non-func "#macro " form) :to-be nil))))
 
+(describe "clojure-match-next-def"
+  (it "handles vars with metadata"
+    (with-clojure-buffer "
+(def ^Integer a 1)
```
vemv, line 66: Some other test cases that should be analog to this one:

* `(def ^:a a 1)`
* `(def ^::a a 1)`
* `(def ^::a/b a 1)`

vemv (commented):
Looking reasonable!

Some suggestions for code coverage, will eye the impl again afterwards

test/clojure-mode-syntax-test.el
OknoLombarda: Done. I've also added `(def ^{:macro true} a 1)`

test/clojure-mode-syntax-test.el
OknoLombarda: Done

clojure-mode.el
OknoLombarda: Done

vemv:
Great work!

I verified that if one discards the fix but keeps the tests, said tests indeed loop forever: https://app.circleci.com/pipelines/github/clojure-emacs/clojure-mode/261/workflows/9315c32d-6ac0-40e0-a7ef-e16eb51218d2/jobs/899

PR #626, opened 2022-07-29
[Fix #625] Fix imenu displaying metadata instead of var name 

### LINKED ISSUE #625: imenu is broken after commit `54a62cc02a0901400a1a3660182a4c898626116d`
## Expected behavior

When I execute `M-x imenu` or any enhanced version of it (like `helm-imenu`) I expect to see list of function names:

image

## Actual behavior

If definition has some metadata it's rendered in the imenu instead of definition name:

image

1. Create a new buffer with clojure-mode
2. Type the following:
```clojure
(def ^:private first-var "Hello")

(def ^{:private true} second-var "World")

(defn some-func
  []
  (print (str first-var second-var)))
```
3. Execute `M-x imenu`

*Note:* Reverting https://github.com/clojure-emacs/clojure-mode/commit/54a62cc02a0901400a1a3660182a4c898626116d fixes the issue.

### clojure-mode version

*`M-x clojure-mode-display-version` returns the following:*

```
clojure-mode (version nil)
```

### Emacs version

*GNU Emacs 29.0.50 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.34, cairo version 1.17.6) of 2022-07-14*

### Operating system

*Linux 5.18.14-arch1-1 #1 SMP PREEMPT_DYNAMIC Sat, 23 Jul 2022 11:46:17 +0000 x86_64 GNU/Linux*

OknoLombarda:
> Fixes https://github.com/clojure-emacs/clojure-mode/issues/625

Fix incorrect bounds for var name being captured in `clojure-match-next-def` when metadata is present in definition. [Context](https://github.com/clojure-emacs/clojure-mode/issues/625#issuecomment-1198600423)

-----------------

Thanks!

### DIFF
--- a/test/clojure-mode-syntax-test.el
+++ b/test/clojure-mode-syntax-test.el
@@ -89,7 +89,31 @@
         (with-clojure-buffer (concat form some-sexp)
           (end-of-buffer)
           (clojure-match-next-def)
-          (expect (looking-at "(def")))))))
+          (expect (looking-at "(def"))))))
+
+  (it "captures var name"
+    (let ((var-name "some-name"))
+      (dolist (form '("(def %s 1)"
+                      "(def %s)"
+                      "(def ^:private %s 2)"
+                      "(def ^{:private true} %s 3)"))
+        (with-clojure-buffer (format form var-name)
+          (end-of-buffer)
+          (clojure-match-next-def)
+          (cl-destructuring-bind (name-beg name-end) (match-data)
+            (expect (string= var-name (buffer-substring name-beg name-end))))))))
+
+  (it "captures var name with dispatch value for defmethod"
+    (let ((name "some-name :key"))
+      (dolist (form '("(defmethod %s [a])"
+                      "(defmethod ^:meta %s [a])"
+                      "(defmethod ^{:meta true} %s [a])"
+                      "(defmethod %s)"))
+        (with-clojure-buffer (format form name)
+          (end-of-buffer)
+          (clojure-match-next-def)
+          (cl-destructuring-bind (name-beg name-end) (match-data)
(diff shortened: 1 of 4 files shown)


bbatsov:
Looks good. Thanks for the quick fix!

test/clojure-mode-syntax-test.el
```diff
@@ -89,7 +89,31 @@
...
+    (let ((var-name "some-name"))
+      (dolist (form '("(def %s 1)"
+                      "(def %s)"
+                      "(def ^:private %s 2)"
+                      "(def ^{:private true} %s 3)"))
+        (with-clojure-buffer (format form var-name)
+          (end-of-buffer)
+          (clojure-match-next-def)
+          (cl-destructuring-bind (name-beg name-end) (match-data)
+            (expect (string= var-name (buffer-substring name-beg name-end))))))))
```
vemv, line 104: I would much appreciate if the expectation wasn't dynamic. Otherwise it is impossible to tell a correct test from a false negative.

Would you kindly offer a follow up PR?

bbatsov: Well, it's not really dynamic, though - it's always "some-name". I get your point, but I'm not sure it's worth changing this.

vemv: You're right, I read the code a little too quickly!

Maybe `string= var-name` can be changed to `string= "some-name"` - this seems one of those cases where clarity is preferable over DRY.

One doesn't always bother learning others' variable names.

OknoLombarda: Created a [PR](https://github.com/clojure-emacs/clojure-mode/pull/627)

PR #627, opened 2022-07-29
Use constant values instead of variables in clojure-match-next-def tests

OknoLombarda:
https://github.com/clojure-emacs/clojure-mode/pull/626#discussion_r933003064


vemv:
Thanks!

PR #629, opened 2022-08-24
Set add-log-current-defun-function

deejayem:
A number of emacs' programming modes[1] set the variable `add-log-current-defun-function` to a function which returns the current defun/function/variable/etc name, e.g. in `lisp/emacs-lisp/lisp-mode.el`: `(setq-local add-log-current-defun-function #'lisp-current-defun-name)` (https://git.savannah.gnu.org/cgit/emacs.git/tree/lisp/emacs-lisp/lisp-mode.el#n683).

This variable is used by which-function-mode to get the name of the current function (otherwise it falls back to some logic based on imenu). It is also used by easy-kill, which then allows the copying of the current defun name (`M-w D`). The latter is the reason I wanted this, as it doesn't work without setting the variable. I haven't discovered anything else that uses it.

The function I have added (`clojure-current-defun-name`) is a copy of lisp-current-defun-name from `lisp-mode.el` (https://git.savannah.gnu.org/cgit/emacs.git/tree/lisp/emacs-lisp/lisp-mode.el#n730), with the small addition of the code to skip metadata (indicated the "Skip metadata" comment), and a tweak to the comments.

I have been using this for a few weeks without issue. It works with def, defn, defn-, defmacro and ns fine.

The convention seems to be to call the function `foo-current-defun-name`, even when the language doesn't use `defun` (e.g. `perl-current-defun-name`, `sh-current-defun-name`, `html-current-defun-name`), so I have stuck with that.

[1] A quick search in `lisp/` in the emacs source turned up these: autoconf, idlwave, cfengine, fortran, cperl-mode, scheme, m4-mode, make-mode, perl-mode, sh-script, tcl, cc-mode, python, octave, ruby-mode, f90, texinfo, sgml-mode, css-mode, tex-mode, diff-mode, org-mode.

-----------------

Thanks!

### DIFF
--- a/clojure-mode.el
+++ b/clojure-mode.el
@@ -525,33 +525,35 @@ replacement for `cljr-expand-let`."
     (advice-add 'paredit-convolute-sexp :after #'clojure--replace-let-bindings-and-indent)))
 
 (defun clojure-current-defun-name ()
-    "Return the name of the defun at point, or nil."
-    (save-excursion
-      (let ((location (point)))
-        ;; If we are now precisely at the beginning of a defun, make sure
-        ;; beginning-of-defun finds that one rather than the previous one.
-        (or (eobp) (forward-char 1))
-        (beginning-of-defun)
-        ;; Make sure we are really inside the defun found, not after it.
-        (when (and (looking-at "\\s(")
-		   (progn (end-of-defun)
-			  (< location (point)))
-		   (progn (forward-sexp -1)
-			  (>= location (point))))
-	  (if (looking-at "\\s(")
-	      (forward-char 1))
-	  ;; Skip the defining construct name, e.g. "defn" or "def".
-	  (forward-sexp 1)
-	  ;; The second element is usually a symbol being defined.  If it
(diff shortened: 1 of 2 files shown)


clojure-mode.el
```diff
@@ -524,6 +524,35 @@ replacement for `cljr-expand-let`."
                  #'clojure-space-for-delimiter-p)
     (advice-add 'paredit-convolute-sexp :after #'clojure--replace-let-bindings-and-indent)))
 
+(defun clojure-current-defun-name ()
+    "Return the name of the defun at point, or nil."
```
bbatsov, line 528: The indentation of the entire function seems off (it's 2 spaces too indented), so you might want to use `indent-region` on it. 

I'd also suggest expanding the docstring to mention this function is meant to be used by `which-func`.

bbatsov:
Hmm, somehow I've totally missed this. Your patch looks good, sans the indentation remark I've made inline.

bbatsov:
Thanks!

PR #630, opened 2022-08-29
[Fix #377] Fix font locking for def forms

### LINKED ISSUE #377: Def highlighting is out of control
## Expected behavior

Proper syntax highlighting
## Actual behavior

image

Write a form containing a word starting with `def` anywhere, e.g. `(defface 'angry-face)`

I get that the intent was to highlight `defresource` etc the same way as `defn` but this really isn't working as intended.  All calls to functions like `(default-user-settings)` are highlighted like a def now, no matter where in the source they appear.

OknoLombarda:
> Fixes https://github.com/clojure-emacs/clojure-mode/issues/377

Well, the solution is mostly described [here](https://github.com/clojure-emacs/clojure-mode/issues/377#issuecomment-409255358), I went with a `defcustom` approach. Also I noticed that font locking is applied multiple times for the same match. I'm not sure if this was intentional, but I just removed it and made a single top level capture for all def forms and everything seems to work. There were tests that checked if custom def forms were properly highlighted, but now they're redundant, so I removed them

-----------------

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Bugs fixed
 
 * [#581](https://github.com/clojure-emacs/clojure-mode/issues/581): Fix font locking not working for keywords starting with a number.
+* [#377](https://github.com/clojure-emacs/clojure-mode/issues/377): Fix everything starting with 'def' being highlighted as a def form.
 
 ## 5.15.1 (2022-07-30)
 
--- a/README.md
+++ b/README.md
@@ -305,11 +305,15 @@ conservative and minimalistic.
 Precise font-locking requires additional data that can obtained from a running
 REPL (that's how CIDER's [dynamic font-locking](https://docs.cider.mx/cider/config/syntax_highlighting.html) works) or from static code analysis.
 
-When it comes to definitions, `clojure-mode` employs a simple heuristic and will treat every symbol named `def`something as a built-in keyword. Still, you'll need to
-teach `clojure-mode` manually how to handle the docstrings of non built-in definition forms. Here's an example:
+When it comes to non built-in definitions, `clojure-mode` needs to be manually instructed how to handle the docstrings and highlighting. Here's an example:
 
 ``` emacs-lisp
 (put '>defn 'clojure-doc-string-elt 2)
+
+(font-lock-add-keywords 'clojure-mode
+                        `((,(concat "(\\(?:" clojure--sym-regexp "/\\)?"
+                                    "\\(>defn\\)\\>")
+                           1 font-lock-keyword-face)))
 ```
 
 **Note:** The `clojure-doc-string-elt` attribute is processed by the function `clojure-font-lock-syntactic-face-function`.
--- a/clojure-mode.el
+++ b/clojure-mode.el
@@ -825,7 +825,37 @@ any number of matches of `clojure--sym-forbidden-rest-chars'."))
 
 (defconst clojure-font-lock-keywords
   (eval-when-compile
-    `( ;; Top-level variable definition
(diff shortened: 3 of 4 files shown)


vemv:
Personally I feel that there are a lot of `def`-like forms, and consequently the default experience will be worse than currently.

Therefore I'd suggest an alternative approach: default to the current approach, but blacklist a few words from the english: `default`, `defer`. There cannot be so many of them.

wdyt?

Cheers - V

vemv:
Is this something that CIDER already enhaces anyway? (or that it could, after this PR)

bbatsov:
> Is this something that CIDER already enhaces anyway? (or that it could, after this PR)

It does, that's one more reason why I prefer the approach suggested in this PR. :-) It will pick up all definitions, macros, etc. (you can turn this off if you don't care about it) That's why it's better to have less assumptions in `clojure-mode` IMO. The smarts should be at the level of CIDER, clojure-lsp, not in `clojure-mode`.

OknoLombarda:
I thought that I'll have at least a week to fix everything before you guys notice PR, but it's like you were waiting for it :D

Also, I think the problem with CI is because of the added variable. Regexp is evaluated at compile time, so it probably can't use a dynamic variable. I have no idea how to fix this at the moment. I'll surely make a research, but if you could give a direction, it would be really appreciated

bbatsov:
Hmm, I thought that `defcustoms` should be available at complile time. Anyways, no rush. 

> I thought that I'll have at least a week to fix everything before you guys notice PR, but it's like you were waiting for it :D

That's how we roll! 😉

OknoLombarda:
Managed to fix it by wrapping `defcustom` form in `eval-and-compile`, but def forms added after compilation are just ignored. Everything works fine if I remove `eval-when-compile` from `clojure-font-lock-keywords`, though
https://github.com/clojure-emacs/clojure-mode/blob/0af29f47eb2ea534eb5e293ade9946795f2e2a01/clojure-mode.el#L826-L827
But I guess compiling this regexp is necessary

OknoLombarda:
So I just excluded regexp responsible for def forms from `eval-when-compile`

bbatsov:
Looking at the [Emacs docs](https://www.gnu.org/software/emacs/manual/html_node/elisp/Eval-During-Compile.html) it seems that `eval-when-compile` is just a performance optimization here. Not sure how big of an effect it has - frankly, even though I've been using Emacs almost 20 years I keep forgetting things about byte-compilation optimizations.

OknoLombarda:
From the [docs](https://www.gnu.org/software/emacs/manual/html_node/elisp/Eval-During-Compile.html):
> The result of evaluation by the compiler becomes a constant which appears in the compiled program.

It seems it's kinda like `^:const` in Clojure. Expression is evaluated and used as a constant

OknoLombarda:
So, should I leave it like this, remove `eval-when-compile` or try to find some other solution?

bbatsov:
I'm wondering about this myself. It'd be simplest to just add the defs to the big form, but that would make them harder to customize. 

Another simple solution would to have two `font-lock-add-keywords` invocations in the mode setup. That's how `clojure-mode-extra-font-locking` works:

```
(font-lock-add-keywords 'clojure-mode
                        `((,(concat "(\\(?:\.*/\\)?"
                                    (regexp-opt clojure-built-in-vars t)
                                    "\\>")
                           1 font-lock-builtin-face)))
```

I'm leaning towards just inlining all the known forms in the existing regexp and adding some snippet in the README about how people can add more themselves.

OknoLombarda:
Done

bbatsov:
Looks good. Thanks!

PR #632, opened 2022-09-03
Fix second element of forms starting with 'def' being highlighted as function name

OknoLombarda:
Follow up to https://github.com/clojure-emacs/clojure-mode/pull/630

Forgot to make function definition regexp only look for `defn` forms. This PR will fix such behaviour:
image

-----------------

Thanks!

### DIFF
--- a/test/clojure-mode-font-lock-test.el
+++ b/test/clojure-mode-font-lock-test.el
@@ -776,17 +776,17 @@ DESCRIPTION is the description of the spec."
      (6 42 clojure-keyword-face)))
 
   (when-fontifying-it "should handle namespaced defs"
-    ("(_c4/defn bar [] nil)"
-     (2 4 font-lock-type-face)
-     (5 5 nil)
-     (6 9 font-lock-keyword-face)
-     (11 13 font-lock-function-name-face))
+    ("(clojure.core/defn bar [] nil)"
+     (2 13 font-lock-type-face)
+     (14 14 nil)
+     (15 18 font-lock-keyword-face)
+     (20 22 font-lock-function-name-face))
 
-    ("(clo/defrecord foo nil)"
-     (2 4 font-lock-type-face)
-     (5 5 nil)
-     (6 14 font-lock-keyword-face)
-     (16 18 font-lock-function-name-face))
+    ("(clojure.core/defrecord foo nil)"
+     (2 13 font-lock-type-face)
+     (14 14 nil)
+     (15 23 font-lock-keyword-face)
+     (25 27 font-lock-type-face))
 
     ("(s/def ::keyword)"
      (2 2 font-lock-type-face)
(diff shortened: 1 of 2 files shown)


OknoLombarda:
Maybe it's better to also remove optional namespace from regexp?

https://github.com/clojure-emacs/clojure-mode/blob/68e0e1355cd21e0752a24432afa091ada4b35b8e/clojure-mode.el#L883

bbatsov:
Yeah, that'd be better. It used to accept all namespaces just because it could be anything.

OknoLombarda:
Done

test/clojure-mode-font-lock-test.el
```diff
@@ -780,13 +780,13 @@ DESCRIPTION is the description of the spec."
      (2 4 font-lock-type-face)
      (5 5 nil)
      (6 9 font-lock-keyword-face)
-     (11 13 font-lock-function-name-face))
+     (11 13 nil))
```
bbatsov, line 783: I guess this example should be updated to use `clojure-core`, otherwise it's kind of pointless.

test/clojure-mode-font-lock-test.el
```diff
@@ -780,13 +780,13 @@ DESCRIPTION is the description of the spec."
...
 
     ("(clo/defrecord foo nil)"
      (2 4 font-lock-type-face)
      (5 5 nil)
      (6 14 font-lock-keyword-face)
-     (16 18 font-lock-function-name-face))
+     (16 18 nil))
```
bbatsov, line 789: Seems to me this example needs to be updated as well to use `clojure.core`, as I doubt they would be a different ns for `defrecord`. Also we should probably make `defrecord` and `deftype` use `font-lock-type-face` for the name of the type.

bbatsov:
It's good we have so many tests - they helped me spot to weird behaviors currently that it would be nice to address.

test/clojure-mode-font-lock-test.el
OknoLombarda: Yeah, makes sense. Done

test/clojure-mode-font-lock-test.el
OknoLombarda: Replaced namespace with `clojure.core`.

> Also we should probably make defrecord and deftype use font-lock-type-face for the name of the type.
We already do, I didn't change that. I changed the test to check if this face is used

bbatsov: Isn't 16-18 the name of the type `foo`? It seems it has no font-locking now.

bbatsov: Ah, I see this wasn't working because of the wrong namespace. My bad.

bbatsov:
Looks good now. Thanks!

PR #638, opened 2022-10-26
Fix imenu with Clojure code in string or comment

dakra:
Ignore error that's thrown from `down-list` when called
with point inside a string or comment.

E.g. with code like:

```
(defn foo []
  (let [a "
(defn b [_]
  (bar {:bla \"bla\"}))"]))
```

`clojure-match-next-def` calls `down-list` with point inside
the string and `down-list` will throw an user-error with
"This command doesn't work in strings or comments".

-----------------

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 
 * [#581](https://github.com/clojure-emacs/clojure-mode/issues/581): Fix font locking not working for keywords starting with a number.
 * [#377](https://github.com/clojure-emacs/clojure-mode/issues/377): Fix everything starting with the prefix 'def' being highlighted as a definition form. Now definition forms are enumerated explicitly in the font-locking code, like all other forms.
+* [#638](https://github.com/clojure-emacs/clojure-mode/pull/638): Fix imenu with Clojure code in string or comment.
 
 ## 5.15.1 (2022-07-30)
 
--- a/clojure-mode.el
+++ b/clojure-mode.el
@@ -777,7 +777,9 @@ Called by `imenu--generic-function'."
       (let (found?
             (deftype (match-string 2))
             (start (point)))
-        (down-list)
+        ;; ignore user-error from down-list when called from inside a string or comment
+        (ignore-errors
+          (down-list))
         (forward-sexp)
         (while (not found?)
           (ignore-errors


dakra:
Friendly ping.
Normally would bump so early but I saw recent activity here and thought this probably just got overlooked by mistake.
If not, all good, take your time :)

bbatsov:
Can you push again? I see something was broken with the CI back then and I guess that's why I didn't review this PR.

clojure-mode.el
```diff
@@ -776,7 +776,8 @@ Called by `imenu--generic-function'."
       (let (found?
             (deftype (match-string 2))
             (start (point)))
-        (down-list)
+        (ignore-errors
```
bbatsov, line 779: I guess some comment here would be useful, otherwise it's really hard to tell why this error needs to be ignored. I'm also a bit worried this might suppress actual errors and complicate the debugging.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 
 * [#581](https://github.com/clojure-emacs/clojure-mode/issues/581): Fix font locking not working for keywords starting with a number.
 * [#377](https://github.com/clojure-emacs/clojure-mode/issues/377): Fix everything starting with the prefix 'def' being highlighted as a definition form. Now definition forms are enumerated explicitly in the font-locking code, like all other forms.
+* [#638](https://github.com/clojure-emacs/clojure-mode/pull/638): Fix imenu with Clojure code in string or comment
```
bbatsov, line 13: Don't forget to terminate the entry with a `.`.

clojure-mode.el
dakra: I looked a bit more into it why I didn't happen before and nobody noticed.
The change to down-list came only in May with this: https://git.savannah.gnu.org/cgit/emacs.git/commit/?id=0b3b295776ce723885c9997ab26d57314db2a5df

I guess a better workaround would be to wrap it in
`unless (ppss-comment-or-string-start (syntax-ppss))` instead of `ignore-errors`,
but `ppss-comment-or-string-start` is only available since Emacs 27 and clojure-mode
is marked to support Emacs 25+.

I updated the commit message with this info and added a short comment.
Happy to rephrase it or change something else.

CHANGELOG.md
dakra: Fixed.

dakra:
CircleCI still red because it couldn't find a usable config.yml

bbatsov:
Ah, CircleCI is killing me these days... Let me try resetting the build token once again. @vemv (or anyone else) - if someone has better ideas what to do about this build issue let me know.

clojure-mode.el
bbatsov: If `ppss-comment-or-string-start` is not complex I guess we can just copy it and inline it.

bbatsov:
@dakra You can try pushing again to see if now the CI will work. I plan to cut a new release soon and it'd be nice for this to make it there.

dakra:
@bbatsov done. but CircleCI instantly failed again :/

bbatsov:
@vemv Any idea what's wrong with CircleCI these days? I'm getting progressively more frustrated with it, given the recent OAuth issues that happened for no apparent reason.

Issue #641, opened 2022-12-03
Support nbb projects


ikappaki:
Support [nbb](https://github.com/babashka/nbb) projects via `nbb.edn` file.

This is required for CIDER to identify the root directory of an `nbb` project. Required for https://github.com/clojure-emacs/cider/pull/3275

Happy to look into the PR.

Thanks

PR #642, opened 2022-12-03
Nbb support

ikappaki:
Add support for nbb including test, addresses #641 

HI @borkdude, I've provisionally added `nbb` in the `intrerpreter-mode-alist` as something that could appear in the shebang so that clojure mode can pick the script as cljs file, but couldn't find any example using this. Is this a valid case?

Thanks

-----------------

Thanks!

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,6 +33,12 @@ jobs:
         entrypoint: bash
     <<: *default-steps
 
+  test-emacs-28:
+    docker:
+      - image: silex/emacs:28-ci-cask
+        entrypoint: bash
+    <<: *default-steps
+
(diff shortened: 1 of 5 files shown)


ikappaki:
I've noticed that emacs-28 was missing from the CI and I took the opportunity to add it, let me know if you would like me to revert.

Btw, whya are the circleci failing on master? some auth key might have expired? Thanks

borkdude:
It might be a valid use case, but I haven't seen it a lot either.

bbatsov:
> Btw, whya are the circleci failing on master? some auth key might have expired? Thanks

I thought I had fixed this already.

bbatsov:
Thanks!

PR #654, opened 2023-08-22
`clojure-find-ns`: never raise errors

vemv:
See also https://github.com/clojure-emacs/cider/issues/2849 for a use case and discussion.

One can see quite a few usages of `clojure-find-ns` in CIDER. Those assume a possible nil value, e.g.:

https://github.com/clojure-emacs/cider/blob/0e76b141054f091fa1a21fea7f2a5604b2f7f8d4/cider-client.el#L129-L130

...so the proposed change would be aligned with the existing usage.

Cheers - V

### DIFF
--- a/clojure-mode.el
+++ b/clojure-mode.el
@@ -2142,19 +2142,19 @@ DIRECTION is `forward' or `backward'."
               (setq candidate (string-remove-prefix "'" (thing-at-point 'symbol))))))))
     candidate))
 
-(defun clojure-find-ns (&optional favor-nil)
-  "Return the namespace of the current Clojure buffer, honor `FAVOR-NIL'.
+(defun clojure-find-ns (&optional suppress-errors)
+  "Return the namespace of the current Clojure buffer, honor `SUPPRESS-ERRORS'.
 Return the namespace closest to point and above it.  If there are
 no namespaces above point, return the first one in the buffer.
 
-If `FAVOR-NIL' is t, errors during ns form parsing will be swallowed,
+If `SUPPRESS-ERRORS' is t, errors during ns form parsing will be swallowed,
 and nil will be returned instead of letting this function fail.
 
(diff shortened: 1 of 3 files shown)


bbatsov:
I'm not opposed to the change, but it is a breaking change regardless. Might be better to either introduce some extra param (e.g. `suppress-errors`) or a different function with a similar name that doesn't raise errors.

vemv:
I find unintentional errors to not quite be an API (they can be considered bugs that are subject to be fixed at any point), however I've applied the feedback 👍

It's green again

clojure-mode.el
```diff
@@ -2142,23 +2142,30 @@ DIRECTION is `forward' or `backward'."
               (setq candidate (string-remove-prefix "'" (thing-at-point 'symbol))))))))
     candidate))
 
-(defun clojure-find-ns ()
-  "Return the namespace of the current Clojure buffer.
+(defun clojure-find-ns (&optional favor-nil)
```
bbatsov, line 2145: I think `suppress-errors` is a better name for this.

vemv:
Done!

clojure-mode.el
```diff
@@ -2142,23 +2142,30 @@ DIRECTION is `forward' or `backward'."
               (setq candidate (string-remove-prefix "'" (thing-at-point 'symbol))))))))
     candidate))
 
-(defun clojure-find-ns ()
-  "Return the namespace of the current Clojure buffer.
+(defun clojure-find-ns (&optional supress-errors)
```
bbatsov, line 2145: It's actually "suppress" (double p). Btw, on second thought "ignore-errors" might be an even better name, but I'm fine with both options.

vemv: Arrgh :)

Amended. I left it at suppress because there's already a construct named `ignore-errors` (probably Elisp would have been fine with it. Humans not so much?)

PR #661, opened 2023-09-04
Fix clojure-find-ns 

p4v4n:
- when ns is preceded by whitespace or inside comment form.
- closes https://github.com/clojure-emacs/clojure-mode/issues/593

### DIFF
--- a/clojure-mode.el
+++ b/clojure-mode.el
@@ -2125,7 +2125,7 @@ content) are considered part of the preceding sexp."
 (make-obsolete-variable 'clojure-namespace-name-regex 'clojure-namespace-regexp "5.12.0")
 
 (defconst clojure-namespace-regexp
-  (rx line-start "(" (? "clojure.core/") (or "in-ns" "ns" "ns+") symbol-end))
+  (rx line-start (zero-or-more whitespace) "(" (? "clojure.core/") (or "in-ns" "ns" "ns+") symbol-end))
 
 (defcustom clojure-cache-ns nil
   "Whether to cache the results of `clojure-find-ns'.
@@ -2153,7 +2153,7 @@ DIRECTION is `forward' or `backward'."
           (save-match-data
             (goto-char end)
             (clojure-forward-logical-sexp)
-            (unless (or (clojure--in-string-p) (clojure--in-comment-p))
+            (unless (or (clojure--in-string-p) (clojure--in-comment-p) (clojure-top-level-form-p "comment"))
               (setq candidate (string-remove-prefix "'" (thing-at-point 'symbol))))))))
     candidate))
 
@@ -2275,6 +2275,8 @@ This will skip over sexps that don't represent objects, so that ^hints and
   (condition-case nil
       (save-excursion
         (beginning-of-defun)
+        (clojure-forward-logical-sexp 1)
+        (clojure-backward-logical-sexp 1)
         (forward-char 1)
         (clojure-forward-logical-sexp 1)
         (clojure-backward-logical-sexp 1)
--- a/clojure-mode.el
+++ b/clojure-mode.el
@@ -2275,11 +2275,7 @@ This will skip over sexps that don't represent objects, so that ^hints and
   (condition-case nil
       (save-excursion
         (beginning-of-defun-raw)
-        ;; Go to beginning of sexp
-        (clojure-forward-logical-sexp 1)
-        (clojure-backward-logical-sexp 1)
         (forward-char 1)
-        ;; Go to beginning of operator
         (clojure-forward-logical-sexp 1)
         (clojure-backward-logical-sexp 1)
         (looking-at-p first-form))
--- a/clojure-mode.el
+++ b/clojure-mode.el
@@ -2275,9 +2275,11 @@ This will skip over sexps that don't represent objects, so that ^hints and
   (condition-case nil
       (save-excursion
         (beginning-of-defun)
+        ;; Go to beginning of sexp
         (clojure-forward-logical-sexp 1)
         (clojure-backward-logical-sexp 1)
         (forward-char 1)
+        ;; Go to beginning of operator
         (clojure-forward-logical-sexp 1)
         (clojure-backward-logical-sexp 1)
         (looking-at-p first-form))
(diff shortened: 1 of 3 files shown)


clojure-mode.el
```diff
@@ -2275,6 +2275,8 @@ This will skip over sexps that don't represent objects, so that ^hints and
   (condition-case nil
       (save-excursion
         (beginning-of-defun)
+        (clojure-forward-logical-sexp 1)
```
bbatsov, line 2278: Perhaps we should add some comments here?

vemv:
Nice work!

The other day I tried to fix a similar issue but my fix broke other things. Could you check if adding these these tests would also pass?

https://github.com/clojure-emacs/clojure-mode/commit/d6a43eef7ca214bdba649d0681b8e8ddc4d9d8f8

I was particularly interested in this case:

```clj
1 (ns foo)
```

however, if it's too hard of a challenge (as it was for me), feel free to leave it.

clojure-mode.el
p4v4n: The change is to handle whitespace before comment form. Not sure if this is the best way to do it though.
Added a couple of comments now to make the intent more clear.

bbatsov: My point was mainly that it looks a bit weird that we're going forward and backwards back-to-back. If that's used often in the code we can make a function I guess.

p4v4n: Agreed. But I am not sure what is the correct abstraction anymore as it's more of a hack.

This is the third occurrence in the codebase but there are other places where similar logic was used with different functions and some places where edge cases are not handled. Even my current change will still break if there are multiple expressions in the same line.

Looks like all these extra steps are only needed in many places because of `beginning-of-defun` behaving 'incorrectly'.
For ex:
```
1 (ns abc) (defn abc []
             (comment
               (println "try (clojure-beginning-of-defun-function) from here"))
             "abc")
```
Now running `(clojure-beginning-of-defun-function)` or `(beginning-of-defun)` from any of the inner expressions takes the cursur to the beginning of the line (before 1).
The expected behaviour IMO from the name would be having the cursor before `(defn`.
The docstring mentions this but there is no way to easy way to control the end location of the cursor.

clojure-mode uses beginning-of-defun a lot so there are a lot of unhandled edge cases.
For example take these two code blocks:
```
(defn s1 [x] (str x)) (+ 1 2)

(defn s2
    [x] (str x))
```
```
(defn s1 [x] (str x)) 
(+ 1 2)
(defn s2
    [x] (str x))
```

Run `M-x cider-eval-defun-at-point`(C-c C-c) inside (+ 1 2) and the result is different in the two scenarios.

p4v4n: tldr: IIUC bad choice of abstraction from `beginning-of-defun` fn has propagated many edge cases possibly in all lisp modes.
Maybe I am missing something basic or someone already solved this issue.

p4v4n: @bbatsov I have raised a separate PR for replacing `beginning-of-defun` fn completely here.
https://github.com/clojure-emacs/clojure-mode/pull/663
After merging that PR these two lines here can be removed.

bbatsov: Thanks for diving deep into this!

p4v4n: Merged latest master and removed the extra lines here.
Will wait for @vemv to resolve the other conversation.

bbatsov:
> A simpler fix would be take your regex change and delete this unlikely test case! Allowing any forms before comment is far more likely scenario than having fake ns in metadata.

I agree.

bbatsov:
Btw, you'll also have to rebase this on top of `master`.

p4v4n:
> Btw, you'll also have to rebase this on top of `master`.

Already merged latest master. The commits should be squashed here so didn't bother with a rebase.

vemv:
@p4v4n thanks much for looking into that!

I'd be fine with no action for the time being. Perhaps, a year from now most people will be using clojure-ts-mode, which since isn't regex-based, shouldn't be as prone to false positives.

test/clojure-mode-util-test.el
```diff
@@ -82,6 +82,21 @@
...
+     (comment
+      (ns bar))"
+      (expect (clojure-find-ns) :to-equal "foo"))
+    (with-clojure-buffer " (comment
+      (ns foo))
+     (ns bar)
+    (comment
+      (ns baz))"
+      (expect (clojure-find-ns) :to-equal "bar"))
+    )
```
vemv, line 99:
```suggestion
      (expect (clojure-find-ns) :to-equal "bar")))
```

test/clojure-mode-util-test.el
```diff
@@ -82,6 +82,21 @@
       (expect (clojure-find-ns) :to-equal "foo"))
     (with-clojure-buffer "(ns ^:bar ^:baz foo)"
       (expect (clojure-find-ns) :to-equal "foo")))
+  (it "should find namespaces with spaces before ns form"
```
vemv, line 85: I'd suggest to also include the following case:

```clj
(comment
  (ns foo))
```

...Asserting that `(clojure-find-ns)` is nil.

That would help us prevent anything like an infinite loop.

vemv (approved):
LGTM after a couple minor suggestions

vemv:
> A simpler fix would be take your regex change and delete this unlikely test case! Allowing any forms before comment is far more likely scenario than having fake ns in metadata.
>> I agree.

When put like this, I actually agree too. (I could as well flip a coin tbh). Feel free to tackle it in a follow-up PR.

p4v4n:
> When put like this, I actually agree too. (I could as well flip a coin tbh). Feel free to tackle it in a follow-up PR.

I was partially joking about deleting the test case. My past experiences with emacs lisp regexes were not very pleasant.(I haven't used the rx macros though TBF)

Will work on it in a separate branch.We can go with the compromise solution if the proper regex is difficult.

test/clojure-mode-util-test.el
p4v4n: Made both of the suggested changes.
Please squash the commits while merging.

PR #663, opened 2023-09-05
Improve support for multiple forms in single lines by replacing `beginning-of-defun` fn

p4v4n:
This change will improve overall support when multiple forms are present on the same line.

- `beginning-of-defun` by default moves the cursor to the beginning of line instead of stopping at the opening paren. This can lead to many edge cases when multiple forms are present in the same line.
- To control this behaviour and stop the cursor at opening paren we need use `beginning-of-defun-function` and `beginning-of-defun-raw` together.

This change replaces all calls to `beginning-of-defun` with `beginning-of-defun-raw`.
(`beginning-of-defun-raw` calls `beginning-of-defun-function` internally which in our case is `clojure-beginning-of-defun-function`)

To see the improvement I have tried M-x cider-eval-defun-at-point(C-c C-c) inside (+ 1 2) in the below example.
```
(defn f1 [x] (inc x)) (+ 1 2)

(defn f2 [x] (str x))
```
Before the change the output is #'user/f2. After this change it correctly evaluates to 3.

### DIFF
--- a/test/clojure-mode-sexp-test.el
+++ b/test/clojure-mode-sexp-test.el
@@ -32,29 +32,36 @@
         ;; make this use the native beginning of defun since this is used to
         ;; determine whether to use the comment aware version or not.
         (expect (let ((beginning-of-defun-function nil))
+                  (clojure-top-level-form-p "comment")))))
+  (it "should return true when multiple forms are present"
+    (with-clojure-buffer-point
+        "(+ 1 2) (comment
+           (wrong)
+           (rig|ht)
+           (wrong))"
+        (expect (let ((beginning-of-defun-function nil))
                   (clojure-top-level-form-p "comment"))))))
 
 (describe "clojure-beginning-of-defun-function"
   (it "should go to top level form"
     (with-clojure-buffer-point
-      "(comment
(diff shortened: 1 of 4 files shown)


bbatsov:
The proposed change looks good to me. Let's just add a couple of unit tests and a changelog entry before we merge it.

test/clojure-mode-sexp-test.el
```diff
@@ -31,30 +31,37 @@
...
+           (rig|ht)
+           (wrong))"
         (expect (let ((beginning-of-defun-function nil))
                   (clojure-top-level-form-p "comment"))))))
 
 (describe "clojure-beginning-of-defun-function"
   (it "should go to top level form"
     (with-clojure-buffer-point
-      "(comment
+      " (comment
```
bbatsov, line 48: Why is the space before `comment` needed?

p4v4n: I modified it(and a couple of other tests) to ensure `begininng-of-defun` default behaviour is no longer present in the code. With `beginning-of-defun` this test would fail with the additional space.

Issue #656, opened 2022-11-05
An extra form before the ns form, on the same line as the ns form, breaks test detection.


enaeher:
## Expected behavior

`cider-test-run-test` runs a test when the point is inside that test.

## Actual behavior

`cider-test-run-test` gives the error "No test at point."

## Steps to reproduce the problem

In a Clojure file containing an `ns` form and one of the forms in `cider-test-defining-forms` (e.g.: `deftest`), add a form before the `ns` form, on the same line. Then navigate the point to one of the test forms and run `cider-test-run-test`.

Here is an example of a namespace which will exhibit the problem:
```clojure
1 (ns some-test
    (:require [clojure.test :refer [deftest is]]))

(deftest unrecognized-test
  (is (= 1 1)))
```

### CIDER version information

```
;; CIDER 1.5.0-snapshot (package: 20220731.522), nREPL 0.9.0
;; Clojure 1.11.1, Java 17.0.3
```

### Lein / Clojure CLI version

Clojure CLI version 1.11.1.1165

### Emacs version

GNU Emacs 28.2 (build 1, aarch64-apple-darwin21.6.0) of 2022-09-13

### Operating system

MacOS Monterey 12.5.1

### JDK distribution

openjdk version "17.0.3" 2022-04-19 LTS
OpenJDK Runtime Environment Corretto-17.0.3.6.1 (build 17.0.3+6-LTS)
OpenJDK 64-Bit Server VM Corretto-17.0.3.6.1 (build 17.0.3+6-LTS, mixed mode, sharing)

vemv:
I gave this a fair shot, but I don't think there's a reasonable fix

Our `clojure-namespace-regexp` is defined as follows:

https://github.com/clojure-emacs/clojure-mode/blob/192a46653fdc27601905f97a044d47764fee1f4e/clojure-mode.el#L2112C1-L2113

It only matches `(ns` if that's how a given line begins (`line-start`). I could remove that, but then other edge cases would begin failing.

It's simpler to recommend users to avoid writing incorrect / dubious code. Formatters and linters can help with that - it can be a good a idea to run them before running tests.

Cheers - V

PR #664, opened 2023-09-07
Fix clojure-find-ns when preceded by other forms

p4v4n:
Closes https://github.com/clojure-emacs/clojure-mode/issues/656
Follow-up for https://github.com/clojure-emacs/clojure-mode/pull/661

- Have initially tried using a regex and syntax table first to detect if namespace is inside metadata but it was too difficult.
- This change re-frames the problem a little. Instead of checking if the namespace is inside string,comment, metadata e.t.c it checks if the current ns form is top-level form.
- Copied most of the test cases from @vemv previous commit.

### DIFF
--- a/clojure-mode.el
+++ b/clojure-mode.el
@@ -57,7 +57,6 @@
 
 ;;; Code:
 
-
 (defvar calculate-lisp-indent-last-sexp)
 (defvar delete-pair-blink-delay)
 (defvar font-lock-beg)
@@ -717,7 +716,7 @@ If JUSTIFY is non-nil, justify as well as fill the paragraph."
             (fill-prefix (clojure-adaptive-fill-function)))
         (do-auto-fill)))))
 
-
+
 ;;; #_ comments font-locking
 ;; Code heavily borrowed from Slime.
 ;; https://github.com/slime/slime/blob/master/contrib/slime-fontifying-fu.el#L186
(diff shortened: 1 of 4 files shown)


clojure-mode.el
```diff
@@ -2270,6 +2270,16 @@ This will skip over sexps that don't represent objects, so that ^hints and
           (backward-sexp 1))
         (setq n (1- n))))))
 
+(defun clojure--is-top-level-form-p (&optional point)
```
bbatsov, line 2274: What's the difference with `clojure-top-level-form-p`? It's a bit confusing to have 2 very similarly named functions IMO.

p4v4n: `clojure--is-top-level-form-p` checks if the sexp following the point is a top level form.
`clojure-top-level-form-p` checks the type of enclosing top level form w.r.t current point.
ex:
`|(ns foo)` -> `(clojure--is-top-level-form-p)` returns true.
`(|ns foo)` -> `(clojure--is-top-level-form-p)` returns nil.
`(comment |(ns foo))` -> `(clojure--is-top-level-form-p)` returns nil.
`(comment (|ns foo))` -> `(clojure-top-level-form-p "commennt")` returns true.

vemv: I'd suggest naming it `clojure--looking-at-top-level-form`

This would be consistent in name and semantics with `looking-at`, I believe.

p4v4n: Renamed the fn to `clojure--looking-at-top-level-form`.

vemv (approved):
Fantastic work, kudos!

PR #670, opened 2023-11-21
Don't highlight vars with colons as keywords

### LINKED ISSUE #653: Symbols with colon in the middle get wrong syntax highlighting
Symbols with colons in the middle (like `a:b`) get separate font faces before and after the colon.

## Expected behavior
Symbols or bindings with a single colon `:` not at beginning or end should be recognized as a single symbol unit for the purpose of syntax highlighting. Specifically, it seems they should get `font-lock-variable-name-face`. 

Per the [Clojure reader reference docs for symbols](https://clojure.org/reference/reader#_symbols):

>Symbols beginning or ending with ':' are reserved by Clojure. A symbol can contain one or more non-repeating ':'s.

## Actual behavior
A symbol with a colon (not at beginning or end) in the name-defining position gets the correct font-face (apparently `font-lock-variable-name-face`), but other instances get `default` for the pre-colon portion and `clojure-keyword-face` thereafter.

This makes every usage of the symbol or binding to get styled like a half-keyword, half-symbol.

Screenshot 2023-07-07 at 01 44 43

```clojure
(ns repl.clojure-mode-font-bug-colon)

(let [foo:bar 1]
  foo:bar)

(def baz:qux 2)

baz:qux

(defn qqq:zzz [] 5)

qqq:zzz
```

### clojure-mode version

```
clojure-mode (version nil)
```

However, jumping to a clojure font-face definition gives me `clojure-mode-20230626.715`. (The clojure-mode invocation was broken in my previous spring 2022 version, too. I installed with package-list-packages from melpa.)

### Emacs version

    GNU Emacs 28.2 (build 1, aarch64-apple-darwin22.1.0, Carbon Version 169 AppKit 2299) of 2022-12-10

### Operating system

macOS Ventura 13.4

daveliepmann:
Fixes #653

Changes syntax highlighting regexp for keywords to match a colon/double-colon only at the beginning of a word, not in the middle. This allows local vars like `foo:bar` to be highlighted correctly instead of like an unknown symbol for the part before the colon and a keyword for the rest.

For my own reference, the "\\<" is per https://www.gnu.org/savannah-checkouts/gnu/emacs/manual/html_node/emacs/Regexp-Backslash.html#Regexp-Backslash

### Test suite
```clojure
(def foo:bar 5)
(def legal*stuff 10)
(def very*****legal 10)

;; wrong
f:bar
foo:bar
defined-in-another-ns/bar:baz
foooooo:bar

;; right
(try)
fubar
inc
legal*stuff
very*****legal
defined-in-another-ns/lorem-ipsum
:baz
::qualified
(:baz {:baz 5})
:qux/foo
::ns/name
*ns*
*assert*
clojure.core/*assert*
(*assert*)
nil
```

-----------------

Thanks!

### DIFF
--- a/clojure-mode.el
+++ b/clojure-mode.el
@@ -1072,12 +1072,13 @@ any number of matches of `clojure--sym-forbidden-rest-chars'."))
       ;; keywords: {:oneword/ve/yCom|pLex.stu-ff 0}
       (,(concat "\\(:\\{1,2\\}\\)\\(" clojure--keyword-sym-regexp "?\\)\\(/\\)"
                 "\\(" clojure--keyword-sym-regexp "\\)")
+       ;; with ns
        (1 'clojure-keyword-face)
        (2 font-lock-type-face)
-       ;; (2 'clojure-keyword-face)
        (3 'default)
        (4 'clojure-keyword-face))
-      (,(concat "\\(:\\{1,2\\}\\)\\(" clojure--keyword-sym-regexp "\\)")
+      (,(concat "\\<\\(:\\{1,2\\}\\)\\(" clojure--keyword-sym-regexp "\\)")
+       ;; without ns
        (1 'clojure-keyword-face)
        (2 'clojure-keyword-face))
 
--- a/test/clojure-mode-font-lock-test.el
+++ b/test/clojure-mode-font-lock-test.el
@@ -265,6 +265,10 @@ DESCRIPTION is the description of the spec."
     ("(colons:are:okay)"
      (2 16 nil))
 
+    ("(some-ns/colons:are:okay)"
+     (2 8 font-lock-type-face)
+     (9 24 nil))
+
     ("(oneword/ve/yCom|pLex.stu-ff)"
      (2 8 font-lock-type-face)
      (9 10 nil)
--- a/test/clojure-mode-font-lock-test.el
+++ b/test/clojure-mode-font-lock-test.el
@@ -722,6 +722,19 @@ DESCRIPTION is the description of the spec."
      (10 10 default)
      (11 30 clojure-keyword-face)))
 
+  (when-fontifying-it "should handle keywords with colons"
+    (":a:a"
+     (1 4 clojure-keyword-face))
+
+    (":a:a/:a"
+     (1 7 clojure-keyword-face))
+
+    ("::a:a"
+     (1 5 clojure-keyword-face))
+
+    ("::a.a:a"
+     (1 7 clojure-keyword-face)))
+
   (when-fontifying-it "should handle very complex keywords"
     (" :ve/yCom|pLex.stu-ff"
      (3 4 font-lock-type-face)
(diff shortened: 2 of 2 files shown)


vemv (commented):
Hi @daveliepmann , this looks like a reasonable addition!

We'd appreciate if unit tests were included.

Thanks - V

daveliepmann:
Hope these suffice

vemv (approved):
LGTM! I'll try it locally

vemv (commented):
The following test cases are valid Clojure so would be worth adding.

One of them would be broken, hopefully the fix won't complicate things?

```
:a:a ;; OK
:a:a/:a ;; the last two characters don't get keyword rendering
::a:a ;; OK
::a.a:a ;; OK
```

daveliepmann:
all 4 seem to work for me — the namespaced keyword path was not touched

vemv:
Looks good. Mind to share a screenshot of the rendering of those 4 cases?

bbatsov:
Out of curiosity - why do people use such symbols in the wild? I know the Clojure parser allows it (I doubt this was intentional, though), but it always seemed like an anti-pattern to me, that's why I never bothered much to support it in `clojure-mode`.

daveliepmann:
> Looks good. Mind to share a screenshot of the rendering of those 4 cases?

Screenshot 2023-11-22 at 10 23 22

The namespace looks wrong on `:a:a/:a`, right?

daveliepmann:
>Out of curiosity - why do people use such symbols in the wild? I know the Clojure parser allows it (I doubt this was intentional, though), but it always seemed like an anti-pattern to me, that's why I never bothered much to support it in clojure-mode.

I started using it on a couple client projects because the lead dev (using Cursive) liked it. We had a part of the codebase which had standardized CRUD structure which cut across namespaces, so it made sense to have query and view vars side-by-side which included both object and verb, e.g. `org:one`, `org:index`, `member:create`. It made sense; the biggest drawback was syntax highlighting on my machine.

Now I mostly use it as a form of lightweight grouping, almost like a pseudo-namespace. E.g. `widget:foo`, `widget:bar`, `screen:baz`. 

I don't see why it would be an anti-pattern. I don't know whether it was originally intentional but it is now explicitly pointed out as legal in the reader reference.

daveliepmann:
> > Looks good. Mind to share a screenshot of the rendering of those 4 cases?
> 
> Screenshot 2023-11-22 at 10 23 22
> 
>~~The namespace looks wrong on `:a:a/:a`, right?~~

I struck out my comment above because on second look, 

1. I don't think I changed this behavior since I didn't touch the namespaced keyword code path, 
2. I don't think it *should* work per `clojure--keyword-sym-forbidden-1st-chars`
3. I think that correctly reflects the reader per "Symbols beginning or ending with ':' are reserved by Clojure." —  the keyword `:a:a/:a` is not valid
4. the corrected version `:a:a/a` works fine:
Screenshot 2023-11-22 at 18 38 51

vemv:
Thanks much!

I don't have a strong opinion for either side at the moment - we can always refine things later, particularly as indeed it's an unrelated concern.

Will cut a stable release.

daveliepmann:
Thanks 🙌

PR #672, opened 2023-11-24
Fix multi-digit lambda args

### LINKED ISSUE #671: Syntax highlighting broken for digits after the first in % args
While working on #670 I noticed that syntax highlighting for `%<n>` args does not match Clojure reader syntax or behavior.

Disclaimer: I acknowledge that `%15` is unidiomatic and this is an insignificant corner case that shouldn't arise in 99.999% of Clojure programming. Nevertheless I'm reporting it because it's a clear bug.

## Expected behavior
Reader-legal numbered percent-style parameters `%1`, `%2`...`%n` to shorthand-syntax anonymous functions `#()` should be syntax-highlighted.

## Actual behavior
Only the first digit of `%<n>` anonymous functions `#()` parameters is syntax-highlighted. Additional digits are not. So `%1`, `%2`, ... `%9` work, but `%10` and further display incorrectly.

Screenshot 2023-11-24 at 08 44 14

It is legal Clojure to provide up to `MAX_POSITIONAL_ARITY` args, per [*Compiler.java*](https://github.com/clojure/clojure/blob/08a2d9bdd013143a87e50fa82e9740e2ab4ee2c2/src/jvm/clojure/lang/Compiler.java#L5448). More than that throws an exception "Can't specify more than 20 params". This matches observed behavior, which clearly supports up to `%20`:

```clojure
(apply #(max %1 %10 %20)
       (take 20 (range)))
;; => 19

(apply #(max %1 %10 %21)
       (take 21 (range)))
;; => syntax error: Can't specify more than 20 params
```

Screenshot 2023-11-24 at 11 59 17

(I don't think it's important or necessarily our job to distinguish between functional % args and syntax-legal ones, especially since the limit is parameterized (see below), so I say we fix `%999` as well as `%19`.)

### clojure-mode version
```
clojure-mode (version 5.18.0)
```

### Emacs version
28.2

### Operating system
macOS 13.5

daveliepmann:
Fixes #671 by syntax-highlighting multi-digit lambda arguments, e.g. `#(+ %10 %19)`. 

Does not attempt to distinguish between %20 (which is both reader-legal and works) and %21 (which is reader-legal but does not compile) because that's complex and depends on Clojure compiler internals.

-----------------

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### Bugs fixed
+
+* [#671](https://github.com/clojure-emacs/clojure-mode/issues/671): Syntax highlighting for digits after the first in % args
+
 ## 5.18.1 (2023-11-24)
 
 ### Bugs fixed
--- a/clojure-mode.el
+++ b/clojure-mode.el
@@ -1058,7 +1058,7 @@ any number of matches of `clojure--sym-forbidden-rest-chars'."))
        1 'clojure-character-face)
       ;; lambda arguments - %, %&, %1, %2, etc
       ;; must come after character literals for \% to be handled properly
-      ("\\<%[&1-9]?" (0 font-lock-variable-name-face))
+      ("\\<%[&1-9]*" (0 font-lock-variable-name-face))
       ;; namespace definitions: (ns foo.bar)
       (,(concat "(\\<ns\\>[ \r\n\t]*"
                 ;; Possibly metadata, shorthand and/or longhand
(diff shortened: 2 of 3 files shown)


bbatsov:
Looks good to me. I guess I never realized people used double-digit lambda args. 😆

daveliepmann:
>I guess I never realized people used double-digit lambda args.

I never have and in 99.9% of situations would advise against them — I just noticed the regex didn't handle legal Clojure while fixing the keyword regex next to it 🤣

PR #678, opened 2024-03-05
Set lexical-binding in clojure-mode-font-lock-test.el

snogge:
Must have been moved to the second line by a well-meaning fill operation. Unfortunately, it has no effect on the second line and the tests broke on buttercup 1.34.

See https://bugs.gentoo.org/926084 as reported in https://github.com/jorgenschaefer/emacs-buttercup/issues/243


bbatsov:
Thanks for spotting and fixing this!

PR #679, opened 2024-03-25
Sync clojure-mode-extra-font-locking.el version with clojure-mode.el

manphiz:
* It looks like the version of clojure-mode-extra-font-locking.el has been out-of-sync with clojure-mode.el and has been stuck in 3.0.0 for a long time.
* This patch sync the version with the current snapshot.

Note: it would be great to fix 5.18.x branch but not really needed I guess.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 ### Bugs fixed
 
 * Fix `clojure-align` when called from `clojure-ts-mode` major mode buffers.
-* [#671](https://github.com/clojure-emacs/clojure-mode/issues/671): Syntax highlighting for digits after the first in `%` args. (e.g. `%10`).
+* [#671](https://github.com/clojure-emacs/clojure-mode/issues/671): Syntax highlighting for digits after the first in `%` args. (e.g. `%10`)
 * [#680](https://github.com/clojure-emacs/clojure-mode/issues/680): Change syntax class of ASCII control characters to punctuation, fixing situations where carriage returns were being interpreted as symbols.
 
 # Changes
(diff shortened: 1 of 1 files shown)


bbatsov:
That's intentional. It's a separate package, so it should have a different version number.

manphiz:
I see.  However, it looks on stable.melpa.org the version of clojure-mode-extra-font-locking is the same as clojure-mode[1].  This may cause an issue that a locally built clojure-mode-extra-font-locking may get superseded by the version on stable.melpa.org, but they are actually the same version.  Is there a way to let stable.melpa.org honor the version specified in the source file?

[1] https://stable.melpa.org/#/clojure-mode-extra-font-locking

bbatsov:
I don't think so, as MELPA uses the newest tag as the version instead of the version metadata. Probably the only solution would be to move `clojure-mode-extra-font-locking` to a separate repo. CIDER's dynamic font-locking made this package somewhat obsolete, so I haven't thought much about it recently.

manphiz:
Ack.  Please feel free to close this (or leave open as you like :)

bbatsov:
@jackrusher For CIDER's eval commands? I don't recall if we ever had font-locking there.

jackrusher:
@bbatsov I must have been hallucinating, but I swear there was syntax highlighting on results in the mini-buffer up until cider 3.0.0 🤔

bbatsov:
Digging in the history of CIDER I see this was the case about 10 years ago, but I'm not sure when and why we changed this. Definitely not related to any changes in `clojure-mode`, though.

bbatsov:
@jackrusher After digging a bit more - seems that's a recent regression caused by https://github.com/clojure-emacs/cider/pull/3661

bbatsov:
And it's now fixed in https://github.com/clojure-emacs/cider/pull/3743 Turned out to be a trivial issue of using the wrong local variable.

jackrusher:
@bbatsov great, thanks!

PR #681, opened 2024-06-03
Don't use custom logic to find source directory of clojure-mode.el

manphiz:
* Or this will prevent buttercup tests from working with installed clojure-mode under ELPA directories.

* A bit more background: in Debian there is autopkgtest which tests the installed package.  dh-elpa enables a mode that runs the same ERT or Buttercup tests against the installed package instead of the source tree.  In this case, it will let the tests pass during build phase, but autopkgtest will fail as this custom code will still try to locate the source code which will no longer be available.

* I have tested under Debian sbuild that removing the code will let the buttercup tests pass under autopkgtest, but not sure whether this is still required for Eldev to work.

-----------------

### DIFF
--- a/test/utils/test-helper.el
+++ b/test/utils/test-helper.el
@@ -25,13 +25,6 @@
 
 (message "Running tests on Emacs %s" emacs-version)
 
-(let* ((current-file (if load-in-progress load-file-name (buffer-file-name)))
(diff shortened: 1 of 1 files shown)


bbatsov:
The CI seems fine, so I guess the change is safe. I see this code is just something I copy-pasted from a different package 10 years ago. (and I don't remember why) 😄

PR #683, opened 2024-12-05
Recognize defn- in `clojure-find-def` for declarations

SophieBosio:
This PR expands on the work of #682.

The problem is that when running `cider-find-def` with point inside the following function,
```clojure
(defn- foo [x] x)
```
it would not recognize `defn-` as the first symbol. Instead it would output `("defn" "-")`. I would expect it to output `("defn-" "foo")`.

This stems from the regex for declarations in `clojure-def-type-and-name-regex`:
https://github.com/clojure-emacs/clojure-mode/blob/63356ee3bd6903e7b17822022f5a6ded2512b979/clojure-mode.el#L2269

It recognizes `def` followed by an entry from the word class `w` or from the symbol class `_`, but it also states that the end must be from the "comment end" class `>`. The hyphen is not included in this class and therefore can't be the final character in the recognized expression.

My suggested fix is to recognize an optional hyphen at the end.

It still recognizes `def-n` and now also `def-n-`

I added a few test cases to the ones added by @frwdrik.

Emacs version:
GNU Emacs 30.0.92 (build 2, aarch64-apple-darwin24.1.0, NS appkit-2575.20 Version 15.1 (Build 24B83)) of 2024-11-01

`clojure-mode` version:
clojure-mode (version 5.20.0-snapshot)

-----------------

Thanks!

### DIFF
--- a/test/clojure-mode-util-test.el
+++ b/test/clojure-mode-util-test.el
@@ -388,7 +388,11 @@
     (with-clojure-buffer-point
         "(defn- |^{:doc \"A function\"} foo [] 1)
          (defn- ^:private bar 2)"
-        (expect (clojure-find-def) :to-equal '("defn-" "foo")))))
+        (expect (clojure-find-def) :to-equal '("defn-" "foo")))
+    (with-clojure-buffer-point
+        "(def-n- |^{:doc \"A function\"} foo [] 1)
+         (defn- ^:private bar 2)"
+        (expect (clojure-find-def) :to-equal '("def-n-" "foo")))))
 
 (provide 'clojure-mode-util-test)
 
--- a/test/clojure-mode-util-test.el
+++ b/test/clojure-mode-util-test.el
@@ -391,9 +391,38 @@
         (expect (clojure-find-def) :to-equal '("defn-" "foo"))))
   (it "should recognize def...-, with or without metadata"
     (with-clojure-buffer-point
+        "(def foo 1)
+         (def- bar| 5)
+         (def baz 2)"
+        (expect (clojure-find-def) :to-equal '("def-" "bar")))
+    (with-clojure-buffer-point
+        "(def foo 1)
+         (deftest- bar |[x y z] z)
+         (def baz 2)"
+        (expect (clojure-find-def) :to-equal '("deftest-" "bar")))
+    (with-clojure-buffer-point
+        "(def foo 1)
+         (defxyz- bar| 5)
+         (def baz 2)"
+        (expect (clojure-find-def) :to-equal '("defxyz-" "bar")))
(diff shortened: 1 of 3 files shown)


test/clojure-mode-util-test.el
```diff
@@ -374,7 +374,25 @@
         "(deftest ^{:a 1} simple-metadata)
          (deftest ^{:a {}} complex-metadata)
          (deftest |no-metadata)"
-        (expect (clojure-find-def) :to-equal '("deftest" "no-metadata")))))
+        (expect (clojure-find-def) :to-equal '("deftest" "no-metadata"))))
+  (it "should recognize defn-"
```
bbatsov, line 378: Probably you should also add a couple of tests showing that anything like `defxyz-` will be accepted, so it's clearer that the `-` will work for every definition form.

test/clojure-mode-util-test.el
```diff
@@ -374,7 +374,25 @@
...
+        (expect (clojure-find-def) :to-equal '("defn-" "bar")))
+    (with-clojure-buffer-point
+        "(def foo 1)
+         (defn- ^:private bar |[x y z] z)"
+        (expect (clojure-find-def) :to-equal '("defn-" "bar")))
+    (with-clojure-buffer-point
+        "(defn- |^{:doc \"A function\"} foo [] 1)
+         (defn- ^:private bar 2)"
+        (expect (clojure-find-def) :to-equal '("defn-" "foo")))
+    (with-clojure-buffer-point
```
bbatsov, line 392: I think this test should be under a different `it` group.

SophieBosio:
> Thanks for the detailed description of the problem and the proposed solution!
> 
> Outside of my small remark about adding some extra tests - your proposed changes look good to me. Also - please mention your fix in the changelog.
> 
> P.S. It's always nice to see a first-time contributor to the project!

My pleasure! It's my first-ever open source contribution, so thank you for the feedback and for the helpful contribution guide!

Please let this newbie know if there's anything else I can improve 😊

bbatsov:
Everything looks great now! Just rebase your branch on top of `master` to address the conflict in the changelog. 

> My pleasure! It's my first-ever open source contribution, so thank you for the feedback and for the helpful contribution guide!

Hopefully the first of many! 🚀

SophieBosio:
> Everything looks great now! Just rebase your branch on top of `master` to address the conflict in the changelog.
> 
> > My pleasure! It's my first-ever open source contribution, so thank you for the feedback and for the helpful contribution guide!
> 
> Hopefully the first of many! 🚀

Glad to hear it! ❤ Thanks to you and to @frwdrik!

Issue #550, opened 2019-12-25
outline-minor-mode doesn't behave correctly with clojure-mode


j-cr:
**TL;DR:** in `clojure-mode-variables`, the line 
`(setq-local outline-regexp ";;;\\(;* [^ \t\n]\\)\\|(")` 
is likely incorrect; replacing it with `(setq-local outline-regexp ";;;;* ")` works for me.

## Expected behavior
Do `M-x outline-minor-mode` in a clojure buffer; the command `outline-insert-heading` should behave correctly.

In particular, call to `outline-insert-heading` on the first line of each pair should result in the second line:

```
;;; heading:
;;; 

;;;; h2:
;;;; 

(foo)
;;; 
```

## Actual behavior
```
;;; heading:
;;; h 

;;;; h2:
;;;; h 

(foo)
( 
```

## Steps to reproduce the problem
See above, tldr; `M-x outline-minor-mode`, then `M-x outline-insert-heading`.

### clojure-mode version information 
clojure-mode (version 5.11.0)

### Emacs version
25.3.1

bbatsov:
Seems something's wrong with the current regexp indeed, although it doesn't seem obviously incorrect. The intention seems to be to check for an extra `;` or some whitespace after `;;;` which makes sense to me.  `\\|(` looks a bit weird and I'm not sure why this was added.

bbatsov:
Btw, it seems we copied this from `lisp-mode`, as there we have:

```cl
  (setq-local outline-regexp ";;;\\(;* [^ \t\n]\\|###autoload\\)\\|(")
```

j-cr:
Re `\\|(`: if I understand correctly, the intent was to treat top-level expressions as headings, so e.g. the visibility of defuns could be cycled as if the were headings. 

But there are packages ([bicycle](https://github.com/tarsius/bicycle/)) that solve this problem by treating top-level exprs explicitly (bicycle knows how to hide them independently of what `outline-regexp` is), so I think it's not a problem really: if the user wants to hide top-level exprs, he can install bicycle or a similar package.

Re `;* [^ \t\n]`: the intention seems to require a space and then some non-whitespace text in order to treat triple semicolon as a heading, thus:

    ;;;                        -- not a heading
    ;;;foo                     -- not a heading
    ;;; foo                    -- heading

but it means that the "space and then a non-whitespace character" part is included in the match, and `outline-insert-heading` uses that match for the heading prefix, thus producing this behavior:

[...]

PR #685, opened 2024-12-09
Restore outline-regexp pattern for top-level forms

### LINKED ISSUE #684: Prior fix to outline-regexp prevents code folding
`outline-regexp` was changed in b1ea6de to address #550. The updated regexp deleted one pattern too many.

**tl;dr:** In order to both fix the issue described in #550 _and_ enable unified comment and code folding using [bicycle][b] (or a similar package), `outline-regexp` should be changed from `;;;;* ` to `;;;;* \\|(`.

`bicycle` uses `outline-regexp` to identify code blocks as well as headings. It's just a light wrapper that unifies two built-in folding packages:`outline-minor-mode` (for comments) and `hideshow` (for code). Removing the alternate pattern `\\|(` in `outline-regexp` prevents `bicycle` from finding top-level code blocks to fold; it only permits comment folding. The alternate pattern to match top-level code blocks can be added back to `outline-regexp` without effecting the fix for `outline-insert-heading`:

```elisp
;; Preserve the fix to issue #550, while enabling code + comment folding
(setq-local outline-regexp ";;;;* \\|(") 
```

[b]: https://github.com/tarsius/bicycle/

_Originally posted by @jeffvalk in https://github.com/clojure-emacs/clojure-mode/issues/550#issuecomment-2526671817_

jeffvalk:
Multiple alternate patterns were previously removed from `outline-regexp` in b1ea6de. This update restores the alternate pattern that enables top-level code forms to participate in `outline-minor-mode`.

Adding this pattern back in does not undo the fix to #550 intended in b1ea6de.

Closes #684

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Bugs fixed
 
 * [cider#3758](https://github.com/clojure-emacs/cider/issues/3758): Improve regexp for clojure-find-def to recognize more complex metadata on vars
+* [#684](https://github.com/clojure-emacs/clojure-mode/issues/684): Restore `outline-regexp` pattern to permit outline handling of top-level forms.
 
 ## 5.19.0 (2024-05-26)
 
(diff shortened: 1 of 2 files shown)


bbatsov:
Thanks for tackling this!

PR #703, opened 2026-06-09
Fix compilation warnings in Emacs 31+

p4v4n:
- Adds a workaround for https://github.com/emacs-eldev/eldev/issues/120
(The issue was with eldev-build.el loading before bytecomp.el and setting eldev--have-byte-compile-warning-function to nil.)
- Fixes all compilations warnings from Emacs and Eldev.

Update: The bytecomp issue is now fixed upstream in 1.11.2. I've updated the PR accordingly.


bbatsov:
Thanks!

